PR #4, opened 2015-03-01
Fix cursor leaks

bhandfast:
Using custom list adapters instead of simple cursor adapters, and loading the data through active android instead of a sql-query.


StephenBlackWasAlreadyTaken:
This looks good! I say merge it!

PR #5, opened 2015-03-04
Dexcom share support

StephenBlackWasAlreadyTaken:
Lots going on here, any eyes on this would be great


tzachi-dar:
Hi,

Can you give us some background here? If I understand correctly share is
the new dexcom system. Do we get raw, data from there or bg from there?
It seems that there have been changes in the calibration algorithm. Are
they related to the share or is this trying to improve the algorithm?

Thanks
Tzachi

StephenBlackWasAlreadyTaken:
Sure, 
So we get ALL info from share, calibrations, calculated values, everything, for now it wont involve the dexdrip algorithm, (I may add it as an option down the line so people can compare the two)
The other changes (from 0 to 1 where there was only a single calibration point) those are just because someone left theirs without calibrating for more than 4 days, then on the next calibration it saw it as just being 1 and set the slope to zero (as it thinks there should never be a time when it has just 1 calibration other then when its only half done processing the double calibration)

tzachi-dar:
I see. I'll try to go over it on this weekend to see if I can see something.

As for one not entering calibrations for 4 days. Well, we should not end
with a slope of 0, but I don't think that we should show something in that
case. We should probably show only something like a message, please enter a
calibration, otherwise we risk showing garbidge.

Tzachi

StephenBlackWasAlreadyTaken:
Anyone trying this out, please note
if using with the share receiver you must enable auth mode 2 in settings, for now

If using kitkat, once disconnected you cant reconnect without having each device forget each other, you can do this on the dex receiver from settings -> share -> forget
on the phone you will want to go into settings -> bluetooth -> click the dexcom and choose forget
Move fast though! if it tries to reconnect while one of them is forgotten but not the other... who knows what will happen!! (really I dont, I havnt tested it, but I will!)

PR #14, opened 2015-05-05
Wake lock and transaction

bhandfast:
Some fixes regarding wake locks and wrapped data saving into a transaction.


StephenBlackWasAlreadyTaken:
Thanks, looks great Merge whenever you are ready

PR #16, opened 2015-05-10
Use less memory to create notifications

mgranberry:
recycle notification bitmaps and use a smaller bitmap for notification tray in order to reduce memory footprint.

Also fixes slope arrow.


StephenBlackWasAlreadyTaken:
Works for me! good catch on the slope arrow again, w should probably do a pass and check to see if we left it around anywhere

merge at will

PR #17, opened 2015-05-11
Snackbar for DB export

mgranberry:
This moves the notification for the file export to a [Snackbar](http://www.google.com/design/spec/components/snackbars-toasts.html) to get rid of a pet peeve of mine, notifications that are a direct result of user action.

It might be good to put error state notifications (device state setup, network errors for mongo, etc) into snackbars too so the user has some idea what's going on when something doesn't work.


app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -309,7 +309,7 @@ public void setSerialDataToTransmitterRawData(byte[] buffer, int len) {
                         sensor.latest_battery_level = transmitterData.sensor_battery_level;
                         sensor.save();
 
-                        BgReading.create(transmitterData.raw_data, this, timestamp);
+                        BgReading.create(transmitterData.raw_data, transmitterData.raw_data, this, timestamp);
```
StephenBlackWasAlreadyTaken, line 312: hu? Is this supposed to be into dexbridge?

edit: yeah it looks like it, you may just want to check this out from master again

StephenBlackWasAlreadyTaken:
Cool! Looks like a good change to me

tzachi-dar:
Can you pass the filtered data instead of the rawdata twice?

mgranberry:
filtered_data is commented out.  I'm going to let this one sit until someone more knowledgeable can recommend a fix.

PR #22, opened 2015-05-19
Dexbridge and pebble support v2

jstevensog:
Dexbridge Support and Pebble off-line Watch face support.
Fixes:
PebbleSync - Duplicate Watch data requests handled in receivedDataHandler().  Not sure what the cause of these multiple requests with the same transactionId are, if it is only my phone or more general in the xDrip code.  Leaving this in for now until it is fully researched.
BgGraphBuilder.unitized_string() - Ensured mmol values are always returned as X.X.  Caused an ERR in the pebble off-line watch face if, say, 11 was returned instead of 11.0.  Due to the less than good coding of unit determination in the pebble watch face, but probably not a bad idea for the Home display anyway.

No other fixes, all other changes are to do with either Dexbridge support or the Pebble support.
Notes:
1) TransmitterData.create(ByteBuffer buffer[], int len, long timestamp) handles Dexbridge packet data (binary as opposed to string).
2) DexCollectionService.setSerialDataToTransmitterRawData(byte[] buffer, int len) handles Dexbridge packet data, and deals with Beacon and Data packets from dexbridge.  Sets Dexcom Transmitter ID from settings if it differs from a received Data or Beacon packet.
3) DexCollectionService.broadcastUpdate(final Stri ng action, final BluetoothGattCharacteristic characteristic) has an "age" on the duplicate packet detection and handling.  Found that sometimes, particularly with Beacon packets, packets could be ignored for way longer than they should.
4) DexCollectionSerice.sendBtMessage(final ByteBuffer message) function added to allow TXID and ACK packets to be sent to Dexbridge.
5) DexCollectionService.convertSrc(final String src) function added to allow the text based Dexcom Transmitter ID/Serial number in settings to be converted to a long id used in Dexcom packets and Dexbridge to filter packets.
6) DexCollectionService.getSrcValue(char ch) function added to encode character to an int.  Used by convertSrc().
7) DexCollectionService.getBridgeBattery() function added to allow the bridge battery value to be gotten (0-100).  This is currently only stored in DexCollectionService, not in mongo/heroku or anywhere else.
8) DexCollectionService.getBridgeBatteryAsString() function added to allow Home to get a formatted string of the Bridge battery level (0-100%)
9) DexCollectionService added Intent("com.eveningoutpost.dexdrip.DexCollectionService.SAVED_BG"), to notify other classes/activities when a new BG is saved.
10) CollectionServiceStarter.isDexbridgeWixel(Context context) added to indicate if DexbridgeWixel is the selected Data source.
11) Home:
   Added area for bridge battery display.
   Added broadcastReciever for com.eveningoutpost.dexdrip.DexCollectionService.SAVED_BG, so it knows when to updateCurrentBgInfo().
12) BgSendQueue added creation of PebbleSync object.
13) Preferences code updated to include the Dexbridge as a data source, and display the "Dexcom Transmitter ID" option when it is selected.


jstevensog:
Closing this pull request due to bug found in BgSendQueue that was creating multiple PebbleSyncs.

PR #26, opened 2015-05-22
Corrected mg/dl bug in bgDelta.

jstevensog:
This corrects a bug in bgDelta that was not correctly determining the status of the units preference, and constantly sending the BG Delta with a decimal place.   The watch app is pretty stupid and uses the presence of a decimal place in either the BG or the Delta as indication of the units being mmol/l.

Pretty simple typo really, and I had corrected it a long time back, but it came back.  Anyway, this is just a single line and will make things easier for everyone using mg/dl.

mg/dl display on the pebble also requires a bug fix in the cgm-pebble-offline, which I have made in my repo at https://github.com/jstevensog/cgm-pebble-offline.


StephenBlackWasAlreadyTaken:
Nice!

PR #27, opened 2015-05-23
Open xDrip on click/tap on widget

AdrianLxM:
Open xDrip on click/tap on widget.

Not tested in xDrip-Experimental.
Just tested in xDrip-master on LG L40 (KitKat),


StephenBlackWasAlreadyTaken:
Thanks!!

PR #30, opened 2015-05-24
Migration clean up

jstevensog:
Removes duplicate creations of filtered_data columns in BgReading and TransmitterRawData.
Also removes selected_filtered_data column creation in BgReading.  This is not used.
Removed 27.sql

### DIFF
--- a/app/src/main/assets/migrations/27.sql
+++ b/app/src/main/assets/migrations/27.sql
@@ -1,3 +1,0 @@
-ALTER TABLE TransmitterData ADD COLUMN filtered_data DOUBLE;
-ALTER TABLE BgReadings ADD COLUMN filtered_data DOUBLE;
-ALTER TABLE BgReadings ADD COLUMN selected_filtered_data BOOLEAN;
--- a/app/src/main/assets/migrations/32.sql
+++ b/app/src/main/assets/migrations/32.sql
@@ -1,1 +1,1 @@
-ALTER TABLE BgReadings ADD COLUMN selected_filtered_data BOOLEAN;
+ALTER TABLE BgReadings ADD COLUMN filtered_data DOUBLE;


app/src/main/assets/migrations/27.sql
```diff
@@ -1,3 +0,0 @@
-ALTER TABLE TransmitterData ADD COLUMN filtered_data DOUBLE;
```
StephenBlackWasAlreadyTaken, line 1: This line does need to be in here I believe, we need to make sure at least one of these migrations adds filterd data to the TransmitterData table, I dont mind if you keep it in this file or add it into migration 32, either would be just fine!

jstevensog, line 1: It is already added to 31.sql.  I decided it was easier to just keep the
two later migration files, and therefore 27.sql became redundant.
31 still keeps it as REAL rather than DOUBLE, but I thought that wouldn't
be a big issue.  Can change all REAL to DOUBLE if you prefer.

On Sun, May 24, 2015 at 12:15 PM, Stephen Black notifications@github.com
wrote:

> In app/src/main/assets/migrations/27.sql
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/30#discussion_r30949607
> :
> 
> > @@ -1,3 +0,0 @@
> > -ALTER TABLE TransmitterData ADD COLUMN filtered_data DOUBLE;
> 
> This line does need to be in here I believe, we need to make sure at least
> one of these migrations adds filterd data to the TransmitterData table, I
> dont mind if you keep it in this file or add it into migration 32, either
> would be just fine!
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/30/files#r30949607
> .

## 

John Stevens

StephenBlackWasAlreadyTaken, line 1: the one in 31 is to add it to the BgReadings table, we also need it in the transmitter data table (which is what this particular line Im referencing was)

jstevensog, line 1: 31 is adding it to TransmitterData.  32 is adding it to BgReadings.
According to what I have.  I simply altered 32.sql to remove the creation
of selected_filtered_data to BgReadings, and add in the creation of
filtered_data.

On Sun, May 24, 2015 at 12:33 PM, Stephen Black notifications@github.com
wrote:

> In app/src/main/assets/migrations/27.sql
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/30#discussion_r30949694
> :
> 
> > @@ -1,3 +0,0 @@
> > -ALTER TABLE TransmitterData ADD COLUMN filtered_data DOUBLE;
> 
> the one in 31 is to add it to the BgReadings table, we also need it in the
> transmitter data table (which is what this particular line Im referencing
> was)
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/30/files#r30949694
> .

## 

John Stevens

StephenBlackWasAlreadyTaken, line 1: Ahh I missread that number! Sorry about that!! Merging!

jstevensog:
No Probs.  Easy to do.  Wish they actually used another method.  This is
simple enough, but can be very confusing.  Would have preferred a single
file with lines that indicate precisely which version they were for.  Would
make grouping them together and readability a lot better.

On Sun, May 24, 2015 at 12:40 PM, Stephen Black notifications@github.com
wrote:

> Merged #30
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/30
> .
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/30#event-312469940
> .

## 

John Stevens

PR #31, opened 2015-05-24
Alert fixes4

tzachi-dar:
Fix a few minor issues in alert code.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
@@ -134,7 +134,7 @@ public class AlertType extends Model {
                 return at;
             }
         }
-        if (UnclearTime > 0) {
+        if ((UnclearTime > 0 ) && bg_unclear_readings_alerts) {
             Log.e(TAG_ALERT, "We are in an clear state, but not for too long. Alerts are disabled");
             return null;
         }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
@@ -118,12 +118,21 @@ public class AlertType extends Model {
 
         Boolean bg_unclear_readings_alerts = prefs.getBoolean("bg_unclear_readings_alerts", false);
         Long UnclearTimeSetting = Long.parseLong(prefs.getString("bg_unclear_readings_minutes", "90")) * 60000;
+        Boolean bg_unclear_readings_alerts_highvolume = prefs.getBoolean("bg_unclear_readings_alerts_highvolume", true);
 
         Long UnclearTime = BgReading.getUnclearTime(UnclearTimeSetting);
         AlertType at;
         if (UnclearTime >= UnclearTimeSetting && bg_unclear_readings_alerts ) {
             Log.w("NOTIFICATIONS", "Readings have been unclear for too long!!");
             Notifications.bgUnclearAlert(context);
+            if(bg_unclear_readings_alerts_highvolume) {
+                Log.e(TAG_ALERT, "We are in an unclear state for too long, ring 55 alert");
+                at = get_alert(LOW_ALERT_55);
+                if(at == null) {
+                    Log.wtf(TAG, "ERROR, the 55 alert is missing");
+                }
+                return at;
+            }
         }
         if (UnclearTime > 0) {
             Log.e(TAG_ALERT, "We are in an clear state, but not for too long. Alerts are disabled");
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -651,7 +651,11 @@ public class BgReading extends Model {
             BgReading latest_entry = BgReading.lastNoSenssor();
             ra = 0;
             rb = 0;
-            rc = latest_entry.age_adjusted_raw_value;
+            if (latest_entry != null) {
+                rc = latest_entry.age_adjusted_raw_value;
+            } else {
+                rc = 105;
+            }
 
             save();
         }
@@ -678,7 +682,7 @@ public class BgReading extends Model {
             return String.valueOf(noise);
         }
     }
-
+    
     // Should that be combined with noiseValue?
     private Boolean Unclear() {
         Log.e(TAG_ALERT, "Unclear filtered_data=" + filtered_data + " raw_data=" + raw_data);
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -683,6 +683,56 @@ public class BgReading extends Model {
         }
     }
     
+    /*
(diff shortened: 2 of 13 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -651,7 +651,11 @@ public void find_new_raw_curve() {
             BgReading latest_entry = BgReading.lastNoSenssor();
             ra = 0;
             rb = 0;
-            rc = latest_entry.age_adjusted_raw_value;
+            if (latest_entry != null) {
```
StephenBlackWasAlreadyTaken, line 654: Im struggling to see how this could have ever been null (This gets called just after bgReading.save no? are we calling it from somewhere else now??

Or did we break the lastNoSenssor query?

OR are we creating bg engries without having a sensor (which would be bad and will break a lot of things)

app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
```diff
@@ -118,14 +118,23 @@ public static AlertType get_highest_active_alert(Context context, double bg) {
...
         Boolean bg_unclear_readings_alerts = prefs.getBoolean("bg_unclear_readings_alerts", false);
         Long UnclearTimeSetting = Long.parseLong(prefs.getString("bg_unclear_readings_minutes", "90")) * 60000;
+        Boolean bg_unclear_readings_alerts_highvolume = prefs.getBoolean("bg_unclear_readings_alerts_highvolume", true);
 
         Long UnclearTime = BgReading.getUnclearTime(UnclearTimeSetting);
         AlertType at;
         if (UnclearTime >= UnclearTimeSetting && bg_unclear_readings_alerts ) {
             Log.w("NOTIFICATIONS", "Readings have been unclear for too long!!");
             Notifications.bgUnclearAlert(context);
+            if(bg_unclear_readings_alerts_highvolume) {
```
StephenBlackWasAlreadyTaken, line 128: Wont this wind up alerting twice? Because the notification.bgUnclearAlert(context) also builds and alerts... I thought we replaced this code in favor of that? Was this accidentally checked in maybe?

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -678,6 +682,70 @@ public String noiseValue() {
...
+    /*
+     * This function comes to check weather we are in a case that we have an allert but since things are
+     * getting better we should not do anything. (This is only in the case that the alert was snoozed before.)
+     * This means that if this is a low alert, and we have two readings in the last 15 minutes, and
+     * either we have gone in 10 in the last two readings, or we have gone in 3 in the last reading, we
+     * don't play the alert again, but rather wait for the alert to finish.
+     *  I'll start with having the same values for the high alerts.
+    */ 
+    
+    public static boolean trendingToAlertEnd(Context context, boolean above) {
```
StephenBlackWasAlreadyTaken, line 695: I think this is a great idea, its going to be tough to explain it, but its going to save so many unnecessary alerts from waking people up!

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -138,6 +138,8 @@ public void onCreate(Bundle savedInstanceState) {
             bindPreferenceSummaryToValue(findPreference("units"));
 
             addPreferencesFromResource(R.xml.pref_notifications);
+//            bindPreferenceSummaryToValue(findPreference("bg_alerts_screen"));
```
StephenBlackWasAlreadyTaken: can we get rid of these since they are commented out anyhow

tzachi-dar: Sure, this is my mistake. Removed

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
tzachi-dar, line 654: Here is a call stack of the issue:
E/AndroidRuntime(17595): FATAL EXCEPTION: Thread-31340
E/AndroidRuntime(17595): java.lang.NullPointerException
E/AndroidRuntime(17595):        at com.eveningoutpost.dexdrip.Models.BgReading.find_new_raw_curve(BgReading.java:655)
E/AndroidRuntime(17595):        at com.eveningoutpost.dexdrip.Models.BgReading.perform_calculations(BgReading.java:522)
E/AndroidRuntime(17595):        at com.eveningoutpost.dexdrip.Models.BgReading.create(BgReading.java:296)
E/AndroidRuntime(17595):        at com.eveningoutpost.dexdrip.Services.WixelReader.setSerialDataToTransmitterRawData(WixelReader.java:408)
E/AndroidRuntime(17595):        at com.eveningoutpost.dexdrip.Services.WixelReader.run(WixelReader.java:378)
I/ActivityManager( 2343): Notify an ApplicationCrash

I believe that the problem comes from the fact that if we have a sensor, but no calibration yet, we do not have a calculated_value.
Since lastNoSenssor should checks for calculated_value != 0 it will return null for the first bg reading.
Do you think that we should do another fix for that?
(by the way, I also wonder why I only get this for syntactic testing, and not in real life...)

StephenBlackWasAlreadyTaken, line 654: Not sure what you mean by syntactic testing but in our case here a bg value can never have a null calculated value (it defaults to zero) due to the way sqlite handles it. Maybe that is why you dont see it in real life?

app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
tzachi-dar, line 128: Just wanted to make sure that I'll always be waken from sleep, in such a case. On a tablet, this alerts were not strong enough.
Please note that it has a setting of itself and is disabled by default.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
tzachi-dar, line 695: Yes, trying to make the alerts less disturbing.
Also intend to make the option to ascending alerts soon.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
tzachi-dar, line 654: Syntactic testing is when I let the wixelReader.runFake actually run.

```
public static BgReading lastNoSenssor() {
    return new Select()
            .from(BgReading.class)
            .where("calculated_value != 0")
            .where("raw_data != 0")
            .orderBy("timestamp desc")
            .executeSingle();
}
```

last no sensor will not return anything with calculated_value = 0. and we have such packets in our db, before the first calibration. So, this seems to be the problem in my understanding.

see the code path of 
BgReading create(double raw_data, double filtered_data, Context context, Long timestamp) 
when sensor!=null and calibration == null.
Maybe we should set calculated_value there as well?

StephenBlackWasAlreadyTaken, line 654: In any event, Im all about checking to make sure things are not null, better safe than sorry!

PR #38, opened 2015-06-08
Alert fixes7

tzachi-dar:
Same as alert fixes7, but after rebase and resolving conflicts.


StephenBlackWasAlreadyTaken:
Looks good still, double check that nothing crazy got merged in but I say its a go

PR #43, opened 2015-06-10
SilentAlerts

StephenBlackWasAlreadyTaken:
@tzachi-dar and @swissalpine let me know if this is what you were thinking

A rundown:

When setting alerts you get to choose how long it will be until it will consider re-raising. This value must be at least 1 and less than the snooze amount (otherwise it would be confusing)

Also a silent mode in which notifications will be populated like normal, only no sounds will be made! Not even at 55 (as many share users still only want to hear noise coming from their dexcoms, or so I have been told)

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
@@ -30,14 +30,15 @@ public class AlertPlayer {
     int volumeBeforeAlert;
     int volumeForThisAlert;
     Context context;
-    
+
     final static int ALERT_PROFILE_HIGH = 1;
     final static int ALERT_PROFILE_ASCENDING = 2;
     final static int ALERT_PROFILE_MEDIUM = 3;
     final static int ALERT_PROFILE_VIBRATE_ONLY = 4;
-    
+    final static int ALERT_PROFILE_SILENT = 5;
+
     final static int  MAX_VIBRATING = 2;
-    final static int  MAX_ASCENDING = 5; 
+    final static int  MAX_ASCENDING = 5;
 
 
     public static AlertPlayer getPlayer() {
@@ -75,7 +76,7 @@ public class AlertPlayer {
     }
 
     public synchronized  void Snooze(Context ctx, int repeatTime) {
-        Log.e(TAG, "Snooze called repeatTime = "+ repeatTime);
+        Log.e(TAG, "Snooze called repeatTime = " + repeatTime);
         stopAlert(ctx, false, false);
         ActiveBgAlert activeBgAlert = ActiveBgAlert.getOnly();
         if (activeBgAlert  == null) {
@@ -111,7 +112,7 @@ public class AlertPlayer {
         }
         if(activeBgAlert.ready_to_alarm()) {
             stopAlert(ctx, false, false);
-            
+
             int timeFromStartPlaying = activeBgAlert.getUpdatePlayTime();
             AlertType alert = AlertType.get_alert(activeBgAlert.alert_uuid);
             if (alert == null) {
@@ -173,7 +174,7 @@ public class AlertPlayer {
         Intent intent = new Intent(ctx, SnoozeOnNotificationDismissService.class);
         return PendingIntent.getService(ctx, 0, intent, PendingIntent.FLAG_UPDATE_CURRENT);
     }
-    
+
     static private int getAlertProfile(Context ctx){
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(ctx);
         String profile = prefs.getString("bg_alert_profile", "ascending");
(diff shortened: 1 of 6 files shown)


swissalpine:
In my oppinion that could be a clever solution.
Am 10.06.2015 03:46 schrieb "Stephen Black" notifications@github.com:

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
```diff
@@ -237,21 +242,24 @@ private void Vibrate(Context ctx, AlertType alert, String bgValue, Boolean overr
                 if(overrideSilent) {
                     PlayFile(ctx, alert.mp3_file, volumeFrac);
                 } else {
-                    builder.setSound(Uri.parse(audioPath), AudioAttributes.USAGE_ALARM);
+                    builder.setSound(Uri.parse(audioPath));
                 }
             }
```
tzachi-dar, line 247: I believe that this is not what you meant to do.
In other words, now it will only vibrate in the case of vibrate only, while we want it to vibrate in all casses that are not silent.

Seems to me that in the line:
 } else if (profile == ALERT_PROFILE_VIBRATE_ONLY ) {
the else is not needed.
Also I believe that the 3 lines:
       NotificationManager mNotifyMgr = (NotificationManager) ctx.getSystemService(Context.NOTIFICATION_SERVICE);
        mNotifyMgr.cancel(Notifications.exportAlertNotificationId);
        mNotifyMgr.notify(Notifications.exportAlertNotificationId, builder.build());

should always happen, so there is no need to have them inside an if.

StephenBlackWasAlreadyTaken, line 247: ahhh good catch, will update!

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
```diff
@@ -265,15 +265,20 @@ public void notificationSetter(Context context) {
...
+            if(alert != null) {
+                int time = alert.minutes_between;
+                Calendar calendar = Calendar.getInstance();
+                AlarmManager alarm = (AlarmManager) getSystemService(ALARM_SERVICE);
+                if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.KITKAT) {
+                    alarm.setExact(AlarmManager.RTC_WAKEUP, calendar.getTimeInMillis() + time, PendingIntent.getService(this, 0, new Intent(this, Notifications.class), 0));
+                } else {
+                    alarm.set(AlarmManager.RTC_WAKEUP, calendar.getTimeInMillis() + time, PendingIntent.getService(this, 0, new Intent(this, Notifications.class), 0));
+                }
             }
```
tzachi-dar, line 282: Main issue that I see here is that there is going to be an alert every time there is a reading. So, if you put this on 20 minutes, you will still get the alert every 5 minutes. So maybe we should limit minutes_between to 5 minutes. even in that case, there is a chance that the alerts and the time of the reading will make it like there are two alerts every 5 minutes.
So maybe we should live the timer as it is today on one minute, but before playing anything check that we haven't played anything on the minutes_between?

StephenBlackWasAlreadyTaken, line 282: since it gets added here https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java#L56 

wouldnt it basicly behave the same as a snoozed alert until that time has passed?

tzachi-dar, line 282: you are right in that point.
So only my firsrt comment remains valid.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
StephenBlackWasAlreadyTaken, line 247: updated, let me know what you think

AdrianLxM:
I think this is exactly what @swissalpine needs for his usecase:
- Phone in the backpack (or somewhere else in hearing range but difficult to reach) - hearing the alarms but difficult to dismiss notifications
- Different monitor (Garmin-Sportsband that can pull the last value via ReST) but cannot dismiss Android-Notifications either.

:) Thank you @StephenBlackWasAlreadyTaken !

swissalpine:
Thank you. Tomorow I will find a second to installiert and test it. But
your solution sounds great.
Am 10.06.2015 23:48 schrieb "AdrianLxM" notifications@github.com:

StephenBlackWasAlreadyTaken:
I actually had to revert it as it went a little haywire, I should have another version that...works...tonight!

Issue #50, opened 2015-06-10
large delta on Widget after missed reading


AdrianLxM:
After one missed Reading, the Widget shows the delta for the last 10 minutes.

The watch seems to divides the delta by two to represent a 5 minute period again.

AdrianLxM:
Proposed fix: https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/51

AdrianLxM:
Quoting @StephenBlackWasAlreadyTaken (https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/51#issuecomment-110969192):
_Unfortunately using slope for this is a no go for share users as the slope isnt an actual slope on the share, rather its just the slope arrow turned into a value.
I wouldnt object to difference divided by difference in timestamps over 5 minutes_

NightWatch workes with the slope for the delta. How about the setting:
Share -> xDrip -> NightWatch->Wear ?

AdrianLxM:
From gitter, better placed here I guess to find, "when it's time" or there is a nighscout decision:
about the slope and Share: As far as I understand the code, calculated_value_slope for Share always is 0 as bgReading.slopefromName() always returns 0.
I believe we could just call find-slope() like with wixel-data and have the cacluclated slope being actually calculated.
Android Studio doesn't find any other "write" usage of calculated_value_slope. But the Widget also uses it in other occasions like for showing the arrow. https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/blob/4a79a7824c94b505f1105f00fbfe572488b9fb5a/app/src/main/java/com/eveningoutpost/dexdrip/xDripWidget.java#L87

AdrianLxM:
... already handled by unifying delta.

PR #51, opened 2015-06-11
use slope instead of diff of last readings

AdrianLxM:
Fix for: https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/issues/50


StephenBlackWasAlreadyTaken:
Unfortunately using slope for this is a no go for share users as the slope isnt an actual slope on the share, rather its just the slope arrow turned into a value.
I wouldnt object to difference divided by difference in timestamps over 5 minutes

AdrianLxM:
Hm, ok @StephenBlackWasAlreadyTaken  I'll close the pull request then. this arises another question:

NightWatch workes with the slope for the delta. How about the setting:
Share -> xDrip -> NightWatch->Wear ?

StephenBlackWasAlreadyTaken:
I suppose we should look into how nightscout is deciding on the delta and use that for consistancy, though I do agree having the delta from the last 5 minutes is more useful then having a delta that may span 20 minutes if for some reason you wandered from your phone or wixel

PR #53, opened 2015-06-11
preventing it to sleep before data gets propagated.

AdrianLxM:
Tested for about 40 minutes.
Works for me. :)


AdrianLxM:
Works for me on a clean merge to today's master.

tzachi-dar:
Where do you free the lock?
I believe that you should add it in a try finally block.

AdrianLxM:
In this case the lock simply "frees" itself after 8 seconds.
This is enough time to propagate the data.

It is _no_t a lock that gives an _exclusive_ right to a resource. It just tells the system to not sleep during the next 8 seconds. Other processes may also run.

PR #58, opened 2015-06-14
Filehandling exceptionhandling

AdrianLxM:
Made DatabaseUtil exception-safe
- close the exceptions in finally
- added some Logs

In Home, I just changed onOptionsItemSelected->onPostExecute. The other diff is due to automatic code reformatting/arrangement by Android-Studio (Adding spaces & reorder some fields/methods according to visibility (private/public)).


StephenBlackWasAlreadyTaken:
I know its probably just your de-linter but its quite difficult to look through so much going on in the diff, I would like to do a full de-lint in its own pr because reading a diff that has some delinting and some actual changes is quite painful

StephenBlackWasAlreadyTaken:
After getting past the delinting, everything looks good! Thanks for this!!

PR #59, opened 2015-06-14
Minutes ago

AdrianLxM:
Show "x Minutes ago" in notification-area, if no other notification is set.

As There is a receiver for _ACTION_TIME_TICK_ that calls _updateCurrentBgInfo();_, it should at least update every full minute.

AFAIK the only time, there is a vadil updated reading and the minutes cannot be shown:

_"Possible bad calibration slope, please have a glass of water, wash hands, then recalibrate in a few!"_ can be set to notification area and then the bg is shown (displayCurrentInfo()). In this case, no "x Minutes ago" is shown, just the notification.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -17,6 +17,7 @@ import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.support.v4.app.NotificationCompat;
 import android.support.v4.widget.DrawerLayout;
+import android.util.Log;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
@@ -203,8 +204,10 @@ public class Home extends ActivityWithMenu {
     }
 
     public void updateCurrentBgInfo() {
+        Log.d("Home", "called updateCurrentBgInfo()");
         final TextView notificationText = (TextView)findViewById(R.id.notices);
         notificationText.setText("");
+        notificationText.setTextColor(Color.RED);
         isBTWixel = CollectionServiceStarter.isBTWixel(getApplicationContext());
         isDexbridgeWixel = CollectionServiceStarter.isDexbridgeWixel(getApplicationContext());
         isBTShare = CollectionServiceStarter.isBTShare(getApplicationContext());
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -345,6 +336,10 @@ public void displayCurrentInfo() {
                 currentBgValueText.setPaintFlags(currentBgValueText.getPaintFlags() | Paint.STRIKE_THRU_TEXT_FLAG);
                 dexbridgeBattery.setPaintFlags(dexbridgeBattery.getPaintFlags() | Paint.STRIKE_THRU_TEXT_FLAG);
             } else {
+                if (notificationText.getText().length()==0){
+                    notificationText.setTextColor(Color.WHITE);
+                    notificationText.setText((System.currentTimeMillis()-lastBgreading.timestamp)/(60*1000) + " Minutes ago");
```
StephenBlackWasAlreadyTaken: thoughts on rearanging this a bit so that

(starting on line 328)

``` java
 if ((new Date().getTime()) - (60000 * 11) - lastBgreading.timestamp > 0) {
    notificationText.setText("Signal Missed");
    ...all that other jazz
} else {
     if (notificationText.getText().length() == 0){
        notificationText.setTextColor(Color.WHITE);
    }
}
notificationText.append("\n" + (System.currentTimeMillis() - lastBgreading.timestamp)/(60*1000) + " Minutes ago");
```

that way even if we know the signal has been missed we can get how long ago the last one was

tzachi-dar: By the way, if we are here maybe we can also add the delta from the last
bg-reading?

AdrianLxM:
@StephenBlackWasAlreadyTaken, nice idea, I "stole" it ;)
@tzachi-dar, Good Idea. Should have gone without saying ;)  I'll push a version with deltas soon. I just have to wait, to intentionally miss the signal to test it.

AdrianLxM:
Because of my reformatting in https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/58 it cannot be automatically merged.
I'm going to manually merge it on a rebase and open a new pull request.

AdrianLxM:
New pull request: https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/61

PR #60, opened 2015-06-15
dont raise low alerts on special dexcom values

StephenBlackWasAlreadyTaken:


tzachi-dar:
How do this values make it into our code?

I believe that we should not let them in (to our BG tables), otherwise, we
will have many other problems including but not limited to:
1) raise/fail alerts.
2) arrow calculations and so on.

Thanks
Tzachi

StephenBlackWasAlreadyTaken:
They are from the dexcom receiver, we need to let them in because they let
users know what is going on on the reciever when using share. The
bgGraphBuilder knows about these values and all displayed c
Values should have to be handled by this first (though it seems the
persistent notification value is not) so some updates will need to be made,
I may pull it out into a separate presenter class with its own queries.

PR #61, opened 2015-06-15
Minute ago delta rebased

AdrianLxM:
Now shows "Minutes ago" and "delta" in the notification area.

If there actually is a notification it works on a regular phone. Small-Screen Phone-Watches cut off the delta. (I can live with not having a delta if "signal missed" or similar is shown. It's not of much use in those cases anyways.)
If there is no notification, it fits fins on my ZGSPAX S8.

(Opened a new pull-request with a rebased version, as it had to be manually merged. Closing the old one: https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/59)

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -17,6 +17,7 @@ import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.support.v4.app.NotificationCompat;
 import android.support.v4.widget.DrawerLayout;
+import android.util.Log;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -333,7 +334,15 @@ public void displayCurrentInfo() {
...
-            if (bgGraphBuilder.unitized(estimate) <= bgGraphBuilder.lowMark) {
+            int minutes = (int)(System.currentTimeMillis() - lastBgreading.timestamp) / (60 * 1000);
+            notificationText.append("\n" + minutes + ((minutes==1)?" Minute ago":" Minutes ago"));
+            List<BgReading> bgReadingList = BgReading.latest(2);
+            if(bgReadingList != null && bgReadingList.size() == 2) {
+                // same logic as in xDripWidget (refactor that to BGReadings to avoid redundancy / later inconsistencies)?
+                notificationText.append("\n"
+                        + bgGraphBuilder.unitizedDeltaString(lastBgreading.calculated_value - bgReadingList.get(1).calculated_value));
+            }
+            if(bgGraphBuilder.unitized(estimate) <= bgGraphBuilder.lowMark) {
```
tzachi-dar, line 348: can you use BgReading.getXRecentPoints(2)
This function also checks that the 2 points are from relatively close times, so it will not display delta, if one point is recent and the other is from 3 hours ago.

AdrianLxM:
@tzachi-dar, at this moment I'd rather have the same code/logic for xDripWidget and this for that they behave the same.

It's already on my todo-list for next weekend to refactor that out to a separate method.

At the moment the delta is calculated at least 4 times throughout the project: Home, xDripWidget, for broadcasting(e.g. Nightwatch) with wixel as datasource, for broadcasting with Share as datasource. The latter two are already not consistent amonst each other.

tzachi-dar:
OK, I see.

AdrianLxM:
Missed 3 lines while merging. Now they are in too.

StephenBlackWasAlreadyTaken:
mergin!

PR #64, opened 2015-06-17
Adapt SystemStatus to small screen

AdrianLxM:
If the screen is to small have the label not left to the data but above.


StephenBlackWasAlreadyTaken:
Looks good to me!

AdrianLxM:
thanks - deleted branch

PR #68, opened 2015-06-19
Fixes slopeOrdinal and makes it use the same code for determining arrows

jstevensog:
as every other display code.
Without this, the Pebble is often off compared to the widget, notification, and main display.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/PebbleSync.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/PebbleSync.java
@@ -171,27 +171,16 @@ public class PebbleSync extends Service {
     }
 
     public String slopeOrdinal(){
-        double slope_by_minute = mBgReading.calculated_value_slope * 60000;
-        String arrow = "0";
-        if (slope_by_minute <= (-3.5)) {
-            arrow = "7";
-        } else if (slope_by_minute <= (-2)) {
-            arrow = "6";
-        } else if (slope_by_minute <= (-1)) {
-            arrow = "5";
-        } else if (slope_by_minute <= (1)) {
-            arrow = "4";
-        } else if (slope_by_minute <= (2)) {
-            arrow = "3";
-        } else if (slope_by_minute <= (3.5)) {
-            arrow = "2";
-        } else {
-            arrow = "1";
-        }
-        if(mBgReading.hide_slope) {
-            arrow = "9";
-        }
-        return arrow;
+        String arrow_name = mBgReading.slopeName();
+        if(arrow_name.compareTo("DoubleDown")==0) return "7";
+        if(arrow_name.compareTo("SingleDown")==0) return "6";
+        if(arrow_name.compareTo("FortyFiveDown")==0) return "5";
+        if(arrow_name.compareTo("Flat")==0) return "4";
+        if(arrow_name.compareTo("FortyFiveUp")==0) return "3";
+        if(arrow_name.compareTo("SingleUp")==0) return "2";
+        if(arrow_name.compareTo("DoubleUp")==0) return "1";
+        if(arrow_name.compareTo("9")==0) return arrow_name;
+        return "0";
     }
 }
 


AdrianLxM:
The only difference I see (at half past 5 in the morning) is it before always defaulted to one and now can default to zero if the slope is steeper than 40?

Did it just happen with Share data? Because Stephen recently fixted the slope number here: 
https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/commit/f0822e0f9a0cfbb343ef259e75f755869bdef2c5#diff-a388fc449f81ad2feaa6c0a9849d3997L403

jstevensog:
This change is only in PebbleSync.  It uses BgReading.slopeName, which
guarantees the Pebble watch face gets exactly the same slope as the Widget,
Notification, and Home screen.
Previously, it was using another value that changed every time it was
looked at, and so the arrow would often be different on the watch.
Cheers

On Fri, Jun 19, 2015 at 1:30 PM, AdrianLxM notifications@github.com wrote:

> The only difference I see (at half past 5 in the morning) is it before
> always defaulted to one and now can default to zero if the slope is steeper
> than 40?
> 
> Did it just happen with Share data? Because Stephen recently changed the
> slope here:
> f0822e0#diff-a388fc449f81ad2feaa6c0a9849d3997L403
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/commit/f0822e0f9a0cfbb343ef259e75f755869bdef2c5#diff-a388fc449f81ad2feaa6c0a9849d3997L403
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/68#issuecomment-113357054
> .

## 

John Stevens

AdrianLxM:
I don't see the different value, sorry.
slopeName() is a function that returns the value based on "calculated_value_slope \* 60000" like it was before in PebbleSync.slopeOrdinal().

I'm just curious, as I had/have a similar problem I'd like to solve: unifying the behaviour of the delta on widget/Home/Broadcast to NightWatch, ...

The widget also uses "lastBgreading.calculated_value_slope \* 60000" and calls BgReading.arrow() (yet another method that basically does the same)

But maybe after a view hours of sleep I'll find the difference ;)

jstevensog:
When PebbleSync called mBgReading.calculated_slope_value, it returned a
different value every time (every minute in this case), meaning that
different arrows could be displayed each minute.  When it calls
mBgReading.slopeName(), BgReading always calculates the same value and
slope.  I only found this out when I added the extra logging in to
BgReadiing which was included in the previous PR.  I didn't bother putting
that logging back into this latest PR, but I can if you wish.

I cannot explain why the difference, as I don't know the BgReading code
that well, or how it is used.  But since I made this change, which is
essentially how all the Notifications, Widget, and home screen code works,
they arrows are ALWAYS  in sync.

BTW, the default was always 0.
Cheers

On Fri, Jun 19, 2015 at 2:17 PM, AdrianLxM notifications@github.com wrote:

> I don't see the different value, sorry.
> slopeName() is a function that returns the value based on
> "calculated_value_slope \* 60000" like it was before in
> PebbleSync.slopeOrdinal().
> 
> I'm just curious, as I had/have a similar problem I'd like to solve:
> unifying the behaviour of the delta on widget/Home/Broadcast to NightWatch,
> ...
> 
> The widget also uses "lastBgreading.calculated_value_slope \* 60000" and
> calls BgReading.arrow() (yet another method that basically does the same)
> 
> But maybe after a view hours of sleep I'll find the difference ;)
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/68#issuecomment-113360757
> .

## 

John Stevens

AdrianLxM:
John, please don't get me wrong, I really think it is cool you found a solution.

I just think if is worth investigating, when it looks like the same result should be returned, just calculated with an indirection step, the result differs. 
It might be a race condition or something like that?
the BgReading.calculated_value_slope should not change every minute. It should be set or calculated (depending on collection method) once when the BgReading is created. _hmmm_
Quite amazing that you found a solution, actually.

In the current master, the default is 1. But that shouldnt make any difference - especially not change the result every minute. https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/68/files#diff-fcc4e5394d58f5e8be1d796f74ffa902L189

Have a nice weekend!

jstevensog:
Hi Adrian,
Sorry, if I sounded stressed out, I am not really, just in coding frenzy.
I am definitely not taking you in any bad way.  I guess I really didn't
articulate things well the first time.

I would also like to get to the bottom of why it always returns a different
calculated_slope_value when called outside the class, because it doesn't
make sense to me.  But I decided that since I didn't really understand the
BgReading code, and since everyone else displaying the arrows was using a
function from BgReading, rather than calculating things out again, I
decided to go with this solution.

Reducing code replication is a good thing, right;)

This code simply translates the already calculated slope arrow from the
internal value to the Pebble watch face expected value.  So, using the
calculated_slope_value internal to the class rather than external.  In a
future watch face, I will simply use this value direct and not convert it.

If anyone gets to the bottom of this, I would be really interested in
knowing why calling calculated_slope_value outside the class causes a slope
value to be returned based on the time since the BgReading was captured and
the time it is looked at again, but calling it inside the class doesn't.  I
am guessing because if it is called within the class, it is using the time
saved in the class?  Just clutching at straws really.

Anyway, I have had the last couple of days off since I worked all last
weekend.  It was a good excuse to spend some time on the code and doing
proper PRs and fixing up my local xDrip-E repo.
You have a good one too.
Cheers

On Fri, Jun 19, 2015 at 3:18 PM, AdrianLxM notifications@github.com wrote:

> John, please don't get me wrong, I really think it is cool you found a
> solution.
> 
> I just think if is worth investigating, when it looks like the same result
> should be returned, just calculated with an indirection step, the result
> differs.
> It might be a race condition or something like that?
> the BgReading.calculated_value_slope should not change every minute. It
> should be set or calculated (depending on collection method) once when the
> BgReading is created. _hmmm_
> Quite amazing that you found a solution, actually.
> 
> In the current master, the default is 1. But that shouldnt make any
> difference - especially not change the result every minute.
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/68/files#diff-fcc4e5394d58f5e8be1d796f74ffa902L189
> 
> Have a nice weekend!
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/68#issuecomment-113372730
> .

## 

John Stevens

PR #69, opened 2015-06-19
Added code to set lastdata to null when a connection is lost.

jstevensog:
Prevents spurious loss of packets when they are ignored as "duplicate".

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
@@ -222,6 +222,7 @@ public class DexCollectionService extends Service {
             } else if (newState == BluetoothProfile.STATE_DISCONNECTED) {
                 mConnectionState = STATE_DISCONNECTED;
                 ActiveBluetoothDevice.disconnected();
+                lastdata = null;
                 Log.w(TAG, "onConnectionStateChange: Disconnected from GATT server.");
                 setRetryTimer();
             }


AdrianLxM:
Just out of curiosity again: When does this happen?

In my opinion, even if there were duplicates - after a connection loss and reconnection there should have passed enough time, that it might be the same value, but shouldn't be a real duplicate.
I anyway just saw this once, when the transmitter is disconnected from the sensor.

I tested it and didn't see any problem -> Ok for how I use xDrip :)

jstevensog:
Yeah, I agree that the time constraint on duplicate packet should do the
trick, but I have seen a kind of race condition happen, where once packet
comes in, isn't processed completely, and the next comes in and gets called
duplicate so ignored.  I think it is a GATT issue that might be resolved in
future, but it is a pain now.

Having the lastdata set to null on connection drop seems to prevent this
from occurring.  A even if the first packet isn't processed immediately it
arrives, the next won't be denied as the lastdata variable is null, so
whichever one actually gets processed first will stop the "second" one
being processed, but the data is dealt with.
Cheers

On Tue, Jun 23, 2015 at 9:48 AM, AdrianLxM notifications@github.com wrote:

> Just out of curiosity again: When does this happen?
> 
> In my opinion, even if there were duplicates - after a connection loss and
> reconnection there should have passed enough time, that it might be the
> same value, but shouldn't be a real duplicate.
> I anyway just saw this once, when the transmitter is disconnected from the
> sensor.
> 
> I tested it and didn't see any problem -> Ok for how I use xDrip :)
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/69#issuecomment-114306195
> .

## 

John Stevens

PR #70, opened 2015-06-19
Keeping backstack + abstract instead of RuntimeException

AdrianLxM:
Back-Button won't close whole xDrip-App from Settings/ScanBT/, ... keeping Home on the "back stack".
Back-Button will still close if xDrip-Activity/Home is on top.

Also: Using "abstract" instead of RuntimeException. This will guarantee at compiletime, that the method is overridden, not cause Exceptions on runtime.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/ActivityWithMenu.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/ActivityWithMenu.java
@@ -6,6 +6,7 @@ import android.os.Bundle;
 import android.support.v4.widget.DrawerLayout;
 import android.util.Log;
 
+import com.eveningoutpost.dexdrip.Home;
 import com.eveningoutpost.dexdrip.NavDrawerBuilder;
 import com.eveningoutpost.dexdrip.NavigationDrawerFragment;
 import com.eveningoutpost.dexdrip.R;
@@ -49,7 +50,8 @@ public abstract class ActivityWithMenu extends Activity implements NavigationDra
             startActivity(intent_list.get(position));
             Set<String> categories = getIntent().getCategories();
             getIntent().getAction();
-            if(categories == null || ! (categories.contains("android.intent.category.LAUNCHER"))){
+            //do not close activity if it is the Launcher or "Home".
+            if(categories == null || ! (categories.contains("android.intent.category.LAUNCHER") || getMenuName().equalsIgnoreCase(Home.menu_name))){
                 finish();
             }
         }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/utils/ActivityWithMenu.java
```diff
@@ -44,11 +48,14 @@ protected void onResume(){
     public void onNavigationDrawerItemSelected(int position) {
         if (position != menu_position) {
             startActivity(intent_list.get(position));
-            finish();
+            Set<String> categories = getIntent().getCategories();
+            getIntent().getAction();
+            //do not close activity if it is the Launcher or "Home".
+            if(categories == null || ! (categories.contains("android.intent.category.LAUNCHER") || getMenuName().equalsIgnoreCase(Home.menu_name))){
```
tzachi-dar: I believe that writing it like this will be more clear. Only adding another () before not (no real change though):
if(categories == null || ! ((categories.contains("android.intent.category.LAUNCHER") || getMenuName().equalsIgnoreCase(Home.menu_name))))

AdrianLxM:
Adding more parenthesis, especially (( and )) doesn't lead to more clarity, imho, just to more parenthesis counting. De Morganing it would be the way to go here.

But actually it can be made much simpler :)

tzachi-dar:
the problem is with the ! operator. people don't except that it is such a
high importance.

app/src/main/java/com/eveningoutpost/dexdrip/utils/ActivityWithMenu.java
```diff
@@ -44,11 +48,12 @@ protected void onResume(){
     public void onNavigationDrawerItemSelected(int position) {
         if (position != menu_position) {
             startActivity(intent_list.get(position));
-            finish();
+            //do not close activity if it is the Launcher or "Home".
+            if (!getMenuName().equalsIgnoreCase(Home.menu_name)) {
```
StephenBlackWasAlreadyTaken, line 52: Good catch!

app/src/main/java/com/eveningoutpost/dexdrip/utils/ListActivityWithMenu.java
```diff
@@ -14,7 +14,7 @@
 /**
  * Created by stephenblack on 6/8/15.
  */
-public class ListActivityWithMenu extends ListActivity implements NavigationDrawerFragment.NavigationDrawerCallbacks {
+public abstract class ListActivityWithMenu extends ListActivity implements NavigationDrawerFragment.NavigationDrawerCallbacks {
```
StephenBlackWasAlreadyTaken, line 17: ah, I knew I was missing something here, I should spend some time to learn more about inheritance in java

PR #71, opened 2015-06-19
Delta fixes

jstevensog:
Brings BgGraphBuilder.unitizedDeltaString into line with BgGraphBuilder.unitized_string, and makes sure the delta has a leading 0 for values <1 or >-1 when units are mmol.  Much less confusing, especially for my crappy eyes.


jstevensog:
Hmmm.  Don't know why all these other changes made it in.  Closing again.

PR #73, opened 2015-06-20
Widgetfix

AdrianLxM:
Updating the numbers after registering the OnClickPendingIntent -> no blank widget between first registering the intent and next update (data or time-tick)

Also cleaning the code a bit for easier debuggability:
- passing parameters instead of using global public variables. -> less prone to side effects
- making private what is not accessed from outside.


AdrianLxM:
... and doing both with one `appWidgetManager.updateAppWidget(appWidgetId, views);` will give us both: updated numbers & click behaviour

mgranberry:
Is this ready to merge?  It looks good to me.

AdrianLxM:
Thanks @mgranberry

PR #74, opened 2015-06-20
Sensor status

AdrianLxM:
Adding sensor status to system status

Did some encapsulation on the way.


ghost:
Great idea Adrian, would be fantastic!

Cheers,
Matthias

StephenBlackWasAlreadyTaken:
Great idea!!

PR #77, opened 2015-06-25
Dex Transmitter ID Uppercase

jstevensog:
There was a likelihood that the user could enter the Dex TXID in
lowercase.  Thanks to AdrianLxM for the Preferences code to force all text
entered into this preference to UPPERCASE.  Also added code in
DexCollectionService to ensure the TXID is in uppercase, just to be on the
safe side.


mgranberry:
lgtm

PR #79, opened 2015-06-26
don't crash when doing redundant wakelock releases on dex share

mgranberry:
This was reported by @diabeticgonewild when it crashed on a new device setup.  It crashed when a sync was attempted without an active sensor session, but several other crashes would occur.

The wakelock is being released in the finally {} block.


StephenBlackWasAlreadyTaken:
Thanks!

PR #80, opened 2015-06-26
use URI text entry type in preferences, fix a typo

mgranberry:
Use URI text entry where appropriate.


StephenBlackWasAlreadyTaken:
Nice, never knew this was a thing, Merging!

PR #86, opened 2015-06-29
Taptochange tapable

AdrianLxM:
To change start and end time, you have to tap the time.
Underneath the time there is a text "(tap to change)" that was not tap-able. Now it is.


StephenBlackWasAlreadyTaken:
Great! It has always driven me a little crazy but not enough to change it myself! Thanks!!!

PR #88, opened 2015-07-01
Don't create duplicate uploads in mongo

mgranberry:


StephenBlackWasAlreadyTaken:
Only downside I can see is that the calibration wont send until a new reading comes in OR until 5 minutes from when the last reading came in, which isnt actually a problem, so I say it looks good to me, merge at will!

StephenBlackWasAlreadyTaken:
Merging this, no issues so far and it looks good!

PR #89, opened 2015-07-01
Fix BLE Lockup issue

jstevensog:
Added code in onConnectionStateChange Bluetooth GATT Callback to correctly
close and null mBluetoothGatt when connection state changes to
DISCONNECTED.
Maybe the if(mBluetoothGatt != null) is overkill, but I believe it is best
to be safe rather than sorry.


AdrianLxM:
Works fine so far with xDrip-wixel.

jstevensog:
Great. Look ok to merge?
BTW,  thanks for testing that. I would have asked, but happy you did it anyway. 
Cheers

---- AdrianLxM wrote ----

AdrianLxM:
After a few intentional bt out of ranges, it works fine! (even reconnected better that my adroid wear ;) )
Looks good to merge to me.

StephenBlackWasAlreadyTaken:
Looks good to me! Merge at will!

jstevensog:
Thanks. Shall do today. 

---- Stephen Black wrote ----

PR #91, opened 2015-07-06
Statistics - Buttoncolor for >=Lollipop

AdrianLxM:
Thank you @LadyViktoria and @mgranberry!

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsActivity.java
@@ -128,12 +128,13 @@ public class StatsActivity extends ActivityWithMenu {
     void setButtonColors() {
 
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-            buttonTD.setBackgroundTintList(null);
-            buttonYTD.setBackgroundTintList(null);
-            button7d.setBackgroundTintList(null);
-            button30d.setBackgroundTintList(null);
-            button90d.setBackgroundTintList(null);
-            ColorStateList csl = new ColorStateList(new int[][]{new int[0]}, new int[]{0xFFAA0000});
+            ColorStateList csl = new ColorStateList(new int[][]{new int[0]}, new int[]{0xFF606060});
+            buttonTD.setBackgroundTintList(csl);
+            buttonYTD.setBackgroundTintList(csl);
+            button7d.setBackgroundTintList(csl);
+            button30d.setBackgroundTintList(csl);
+            button90d.setBackgroundTintList(csl);
+            csl = new ColorStateList(new int[][]{new int[0]}, new int[]{0xFFAA0000});
             switch (state) {
                 case TODAY:
                     buttonTD.setBackgroundTintList(csl);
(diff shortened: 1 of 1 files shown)


StephenBlackWasAlreadyTaken:
Hrm, with this my button colors are all always grey, wheras before I had red on the current one (running 5.0.2)

app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsActivity.java
```diff
@@ -123,28 +126,61 @@ public void onPageScrollStateChanged(int state) {
...
-                button7d.getBackground().setColorFilter(new LightingColorFilter(0xFFFFFFFF, 0xFFAA0000));
-                break;
-            case D30:
-                button30d.getBackground().setColorFilter(new LightingColorFilter(0xFFFFFFFF, 0xFFAA0000));
-                break;
-            case D90:
-                button90d.getBackground().setColorFilter(new LightingColorFilter(0xFFFFFFFF, 0xFFAA0000));
-                break;
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
StephenBlackWasAlreadyTaken: shouldnt this be checking the version I am running rather than the build version? (or am I misreading this)

AdrianLxM:
hm, well, I think then we'll have to try to set the one that should be coloured red to `null` first?
https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/91/files#diff-fe40678237a14aeb62879980cf829f27R140 : 
Could you please try and change this line to
`buttonTD.setBackgroundTintList(null);
buttonTD.setBackgroundTintList(csl);`

I have no device to test and on the 2 5.1 devices my friends tested for it worked. There are reported problems with 5.0. :/

app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsActivity.java
StephenBlackWasAlreadyTaken: ignore me here Im misreading things

AdrianLxM:
@StephenBlackWasAlreadyTaken 
It is more or less the version you are running.
It is the build version of the framework, not of the programme.

http://developer.android.com/reference/android/os/Build.VERSION.html#SDK_INT
"The user-visible SDK version of the framework; its possible values are defined in Build.VERSION_CODES."

(And we are already using it in other places like https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java#L163)

But if you tested the earlier version (current master) on 5.0 and it changed the backgrounds, we should check if it is `>= LOLLIPOP_MR1` anyways.

app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsActivity.java
```diff
@@ -123,28 +126,61 @@ public void onPageScrollStateChanged(int state) {
...
+                case D90:
+                    button90d.getBackground().setColorFilter(new LightingColorFilter(0xFFFFFFFF, 0xFFAA0000));
+                    break;
+            }
         }
+
+
+
+
+
```
StephenBlackWasAlreadyTaken: mind removing some of this whitespace

StephenBlackWasAlreadyTaken:
Great! Merging!

PR #93, opened 2015-07-06
Wakelock in onConnectionStateChange

jstevensog:
Added a wakelock in onConnectionStateChange to ensure the BLE GATT is
closed quickly.  Unlikely this is an issue, but just in case.
Wakelock is a maximum 100ms.

Shows some significant improvement in the stability of the service with xBridge.


StephenBlackWasAlreadyTaken:
if it takes more than the time it was aquired for AND hits the finally, wont it crash since its already been released?

jstevensog:
Just copied the code that already existed in other modules. I am no expert on how it works. Hasn't crashed in over 3 weeks.  I have had the most stable data capture with the sleeping xBridge since these two wake locks went in. No crashes.

---- Stephen Black wrote ----

PR #96, opened 2015-07-19
Allow smart alerting (don't alert if things are getting better)

tzachi-dar:


app/src/main/java/com/eveningoutpost/dexdrip/Services/WixelReader.java
```diff
@@ -337,7 +337,7 @@ public void run()
 	        		// Make sure we do not report packets from the far future...
 	        		if ((LastReading.CaptureDateTime > LastReportedTime ) &&
 	        		        (!almostEquals(LastReading, LastReportedReading)) &&
-	        		        LastReading.CaptureDateTime < new Date().getTime() + 12000) {
+	        		        LastReading.CaptureDateTime < new Date().getTime() + 120000) {
```
StephenBlackWasAlreadyTaken, line 340: Makes sense, I think 2 minutes is a good window

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/IdempotentMigrations.java
```diff
@@ -35,12 +36,11 @@ private void migrateBGAlerts() {
                 highMark = highMark * Constants.MMOLL_TO_MGDL;
                 lowMark = lowMark * Constants.MMOLL_TO_MGDL;
             }
-            int bg_snooze = Integer.parseInt(prefs.getString("bg_snooze", "20"));
             boolean bg_sound_in_silent = prefs.getBoolean("bg_sound_in_silent", false);
             String bg_notification_sound = prefs.getString("bg_notification_sound", "content://settings/system/notification_sound");
 
-            AlertType.add_alert(null, "High Alert", true, highMark, true, 1, bg_notification_sound, 0, 0, bg_sound_in_silent, bg_snooze);
-            AlertType.add_alert(null, "Low Alert", false, lowMark, true, 1, bg_notification_sound, 0, 0, bg_sound_in_silent, bg_snooze);
+            AlertType.add_alert(null, "High Alert", true, highMark, true, 1, bg_notification_sound, 0, 0, bg_sound_in_silent, SnoozeActivity.getDefaultSnooze(true));
```
StephenBlackWasAlreadyTaken: Im kind of against this change, if someone previously set their snooze (with the old style alerts) I think we should keep that value. Is there any reason for this change otherwise?

app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
```diff
@@ -465,7 +465,7 @@ public static void testAlert(
             at.end_time_minutes = end_time_minutes;
             at.override_silent_mode = override_silent_mode;
             at.default_snooze = snooze;
-            new AlertPlayer().getPlayer().startAlert(context, at, "TEST");
+            AlertPlayer.getPlayer().startAlert(context, false, at, "TEST");
```
StephenBlackWasAlreadyTaken, line 468: Good call!

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/IdempotentMigrations.java
tzachi-dar: OK, I have changed that. 
If there is an old default, will use it. If there is no default value (a new install) will use the new value.

StephenBlackWasAlreadyTaken:
Looks good! Merge when you are ready!!

PR #97, opened 2015-07-20
Change to new base URI scheme for API uploader

mgranberry:


jasoncalabrese:
nice @mgranberry, this should reduce a lot of confusion in the community and also help people move away from direct mongo uploads and to a real-time event based system.

PR #99, opened 2015-07-20
Don't disable notification graph when alerts are disabled

mgranberry:


StephenBlackWasAlreadyTaken:
Nice, good catch! Merge at will

PR #103, opened 2015-07-20
Cleanup

bhandfast:
Had some trouble reading bit of the code so I tried to make it more clear, mainly using Android Studio's refactoring tools.


StephenBlackWasAlreadyTaken:
Nice, much much cleaner! Thanks!

PR #104, opened 2015-07-21
Hold communication WakeLock longer

mgranberry:
This isn't checking to ensure that it isn't released twice.  It doesn't look like it could be, but it is a bit tricky following all of the inverted control flows here.  Please pay attention during the review.  It doesn't appear to be crashing.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexShareCollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexShareCollectionService.java
@@ -262,14 +262,14 @@ public class DexShareCollectionService extends Service {
     }
 
     public void attemptRead() {
-        PowerManager powerManager = (PowerManager) getApplicationContext().getSystemService(getApplicationContext().POWER_SERVICE);
-        PowerManager.WakeLock wakeLock1 = powerManager.newWakeLock(PowerManager.PARTIAL_WAKE_LOCK,
+        PowerManager powerManager = (PowerManager) getApplicationContext().getSystemService(Context.POWER_SERVICE);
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Services/DexShareCollectionService.java
```diff
@@ -289,6 +289,11 @@ public void call(EGVRecord[] egvRecords) {
                                         if (egvRecords != null) {
                                             Log.d(TAG, "Made the full round trip, got " + egvRecords.length + " EVG Records");
                                             BgReading.create(egvRecords, addativeSystemTimeOffset, getApplicationContext());
+                                            {
```
StephenBlackWasAlreadyTaken, line 292: is this block doing anything?

tzachi-dar, line 292: wouldn't a logic of:
lock
try {

} finaly
{
free lock
}

be more clear, and with a smaller chance to break?

StephenBlackWasAlreadyTaken, line 292: unfortunately @tzachi-dar this is all happening async so a normal try finally wrapping this would cut off the wakelock probably before a single response was received from the share

tzachi-dar, line 292: My mistake.
Looking this at github gives a very bad impression of what is actually going on....
Had to clone the repository to really see it.

StephenBlackWasAlreadyTaken:
Im game for this being merged, we may revisit it down the road but like you said, worse case scenario is it wastes a little extra battery on a checksum error or something

PR #105, opened 2015-07-21
Debugging BLE

bhandfast:
Cleaned up some code while debugging why the BLE device didn't send updates (I use a BLE Mini instead of HM-10).

Turned out that HM-10 is hard-coded now which wasn't the case in previous versions. I'm upgrading from a ~5 months old version.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
@@ -59,7 +59,6 @@ public class DexCollectionService extends Service {
 
     public DexCollectionService dexCollectionService;
 
-    private BluetoothManager mBluetoothManager;
     private BluetoothAdapter mBluetoothAdapter;
     private BluetoothGatt mBluetoothGatt;
     private ForegroundServiceStarter foregroundServiceStarter;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -242,32 +265,39 @@ public void onConnectionStateChange(BluetoothGatt gatt, int status, int newState
...
-            } else {
+            if (status != BluetoothGatt.GATT_SUCCESS) {
                 Log.d(TAG, "onServicesDiscovered received: " + status);
+                return;
+            }
+
+            Log.d(TAG, "onServicesDiscovered received status: " + status);
+
+            final BluetoothGattService gattService = mBluetoothGatt.getService(xDripDataService);
+            if (gattService == null) {
```
StephenBlackWasAlreadyTaken, line 276: Ah I get it now, so if its not finding your device list some logs so you can see what your characteristics are, makes sense!

StephenBlackWasAlreadyTaken:
Lgtm!! Merge at will

StephenBlackWasAlreadyTaken:
(merging, hope you dont mind!)

PR #106, opened 2015-07-21
Alarm parsing numbers

bhandfast:
Fixing a issue when decimals are printed 10,5 (which is my current locale) instead of 10.5.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
@@ -1,19 +1,13 @@
 package com.eveningoutpost.dexdrip;
 
-import java.text.DecimalFormat;
-import java.text.ParseException;
-import java.text.SimpleDateFormat;
-import java.util.Date;
-import java.util.List;
-
 import android.app.AlertDialog;
 import android.app.Dialog;
 import android.app.TimePickerDialog;
 import android.content.Context;
 import android.content.DialogInterface;
-import android.database.Cursor;
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.database.Cursor;
 import android.media.Ringtone;
 import android.media.RingtoneManager;
 import android.net.Uri;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
```diff
@@ -350,30 +357,52 @@ private boolean verifyThreshold(double threshold) {
         return true;
     }
 
+    private double parseDouble(String str) {
```
StephenBlackWasAlreadyTaken, line 360: ahhh I didnt realize you had defined this method, I was wondering how on earth this was suppsoed to work, now I get it

app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
```diff
@@ -350,30 +357,52 @@ private boolean verifyThreshold(double threshold) {
...
-                try {
-                    alertReraise = Integer.parseInt((reraise.getText().toString()));
-                }
-                catch (NumberFormatException nfe) {
-                    Log.e(TAG, "Invalid number", nfe);
-                }
+                Integer alterReraiseInt = parseInt(reraise.getText().toString());
+                if(alterReraiseInt ==null)
+                    return;
+                alertReraise = alterReraiseInt;
```
StephenBlackWasAlreadyTaken, line 404: any reason for the alterReraiseInt existing instead of just setting alertReraise off the bat?

bhandfast, line 404: alertReraise is defined as an int but the parseInt might return null which will cause a NPE.

bhandfast, line 404: alertReraise could perhaps be redefined as an Integer, not sure.

StephenBlackWasAlreadyTaken, line 404: ah, gotcha

StephenBlackWasAlreadyTaken:
so is havint the comma in there a problem for them working, or just looks funny? Why wouldn't we want to keep with the phones locale?

bhandfast:
Actually, the textfields were filled with correct formatting, eg "10,5", but it couldn't be saved since the code expected the fields were formatted as "10.5" regardless of the phones locale. This patch essentially makes the code parse numbers in the phones locale.

StephenBlackWasAlreadyTaken:
ahhhhh gotcha gotcha, Makes sense!

Do you see this being an issue elsewhere? (also, merge at will!)

bhandfast:
I haven't found any more yet. Still looking through all the new things since my upgrade. I noticed that the settings number input works different (and maybe better?) from the alarm activity.

PR #107, opened 2015-07-21
Big Tablet fixes:

tzachi-dar:
1) Don't use a separate xml file for xlarge tablets.
2) Use common code where possible for wifi and bt.
3) print time + bg on the same line for big tablets.


StephenBlackWasAlreadyTaken:
Looks good, thanks for taking that refactor a step further! I think down the line we should pull some things out into the dimens files and reduce the need to do this in the activity code, but thats for down the line!

PR #108, opened 2015-07-22
Make bt scan smarter

StephenBlackWasAlreadyTaken:


app/src/main/java/com/eveningoutpost/dexdrip/BluetoothScan.java
```diff
@@ -155,16 +172,93 @@ public void run() {
...
+            returnToHome();
+        }
+    }
+
     @Override
     protected void onListItemClick(ListView l, View v, int position, long id) {
         Log.d(TAG, "Item Clicked");
         final BluetoothDevice device = mLeDeviceListAdapter.getDevice(position);
-        if (device == null) return;
+        if (device == null || device.getName() == null) return;
```
StephenBlackWasAlreadyTaken, line 252: If we are returning here should I remove the code that lists nameless devices as 'unknown device' and just have it skip them entirely?

PR #109, opened 2015-07-22
config barcode scanner

mgranberry:
Thanks to Kevin Lee and the rest of the android-uploader crew.

This isn't quite ready yet.  I would like to pull in the android-uploader core.jar as a maven dependency, but I haven't yet gotten that sorted out.

I also want to decode the transmitter box serial number, but I'd like to have another example before I decide it's working well enough to include.


mgranberry:
Should be ready to merge whenever you're ready.

StephenBlackWasAlreadyTaken:
Seems awesome!!

PR #115, opened 2015-07-29
Unify deltas

AdrianLxM:
Now using the same method for:
- Pebble
- Widget
- Persistent Notification
- Home-Screen

It will not show a delta (will show "--" instead) if the last two values are more than 20 minutes apart from each other.

The local broadcast to NightWatch now also uses the calculated slope and not the ordinal one. As the delta is calculated within NightWatch and not xDrip, the 20-minutes rule doesn't work here.

I tested it for android-wear and the xDrip-App both in mg/dl and mmol/l.
As I don't have a pebble, \* **It would be nice, if you could test it on the pebble** *. Thanks!


StephenBlackWasAlreadyTaken:
Looks good to me, @mgranberry or @jstevensog any chance you can test with this change

StephenBlackWasAlreadyTaken:
(Specifically using the pebble)

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -539,17 +558,11 @@ public void perform_calculations() {
...
-            double y1 = calculated_value;
-            double x1 = timestamp;
-            double y2 = second_latest.calculated_value;
-            double x2 = second_latest.timestamp;
-            if(y1 == y2) {
-                calculated_value_slope = 0;
-            } else {
-                calculated_value_slope = (y2 - y1)/(x2 - x1);
-            }
+            calculated_value_slope = calculateSlope(this, last_2.get(1));
```
tzachi-dar, line 565: on this function there is no check that the two points are closed to each other. They can both be from one hour ago, or worth, they can be one from now and one from one hour ago.

AdrianLxM:
This is the function that calculates the "calculated_value_slope" that is used in different places throughout the code, not just for the delta shown in xDrip. I wouldn't change this behaviour during a "feature freeze" for xDrip beta 2.0.
My plan for "after 2.0" is to get more into "cleaning" the data structure. But as this slope even slightly touches the core calibration logic (it is used when calculating the calibration confidence that again is used for calculating the weights, ...) I would do this with a bit of caution and not for an "almost live" system like the beta 2.

For the delta String generation, there is a separate check, that both readings are not more than 20 minutes apart from each other:
https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/115/files#diff-d6adfa0389fad4f508f39cc92a694ea2R330

If both are from some time ago but close to each other in time, I see no problem or I'd even prefer to have a delta. The "minutes ago" field will show a high number and the text will be in red and struck through.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
```diff
@@ -327,21 +327,31 @@ public String unitized_string(double value) {
         }
     }
 
-    public String unitizedDeltaString(double value) {
+    public String unitizedDeltaString(boolean showUnit) {
+
+        List<BgReading> last2 = BgReading.latest(2);
+        if(last2.size() < 2 || last2.get(0).timestamp - last2.get(1).timestamp > 20 * 60 * 1000){
+            // don't show delta if there are not enough values or the values are more than 20 mintes apart
+            return "--";
```
StephenBlackWasAlreadyTaken: @jstevensog would the pebble be able to handle this gracefully?

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
```diff
@@ -327,21 +327,31 @@ public String unitized_string(double value) {
...
+        double value = BgReading.currentSlope() * 5*60*1000;
+
         DecimalFormat df = new DecimalFormat("#");
-        df.setMaximumFractionDigits(1);
         String delta_sign = "";
-        if (value > 0.1) { delta_sign = "+"; }
+        if (value > 0) { delta_sign = "+"; }
         if(doMgdl) {
-            return delta_sign + df.format(unitized(value)) + " mg/dl";
+            df.setMaximumFractionDigits(1);
```
StephenBlackWasAlreadyTaken: Its also my understanding that if the pebble sees a single decimal point it assumes its dealing with mmol

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
```diff
@@ -327,21 +327,31 @@ public String unitized_string(double value) {
...
+            return delta_sign + df.format(unitized(value)) +  (showUnit?" mg/dl":"");
         } else {
+            df.setMaximumFractionDigits(2);
```
StephenBlackWasAlreadyTaken: And I believe when the pebble is dealing with mmol it wants just one decimal point

AdrianLxM:
@StephenBlackWasAlreadyTaken: Good point.  @jstevensog wanted to adopt the pebble watchface.
We'll have to wait before merging this pr into the beta 2 to publish both the android app and the watchface simultaneously.

StephenBlackWasAlreadyTaken:
Yeah I double checked the code and its very particular about the values
it's given, for mgdl it can't have a decimal at all, for mmol it needs a
decimal but can only have one digit before and after (or it will just be
zero)

StephenBlackWasAlreadyTaken:
Lets do it! Looks solid to me!

PR #119, opened 2015-07-31
Make sure that navigation drawer intent list and names match each other.

tzachi-dar:
This fixes a bug where the user would choose one intent and another one would open.

To repro this do the following: start a new sensor, wait for two readings and choose double calibration. The StopSensor intent will open.

This happens since the list of intent changes but it was cached in a few places.
This removes the cache from a few places and make sure that on NavigationDrawerFragment intent_list is always updated when menu_option_list changes.


StephenBlackWasAlreadyTaken:
Thanks!! Looks good to me!

PR #120, opened 2015-07-31
Eliminating a brand name

AdrianLxM:


AdrianLxM:
Update: "Data Collection Method" -> "Hardware Data Source" as suggested by @jwedding

StephenBlackWasAlreadyTaken:
Good call!! Thanks!

PR #121, opened 2015-08-02
RestartCollectionService

StephenBlackWasAlreadyTaken:
Use a name of the collection method that is passed in so we are sure we are getting the new value when it is changed in preferences

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CollectionServiceStarter.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CollectionServiceStarter.java
@@ -30,6 +30,8 @@ public class CollectionServiceStarter {
         }
         return false;
     }
+    public static boolean isBTWixel(String collection_method) { return collection_method.equals("BluetoothWixel"); }
+
     public static boolean isDexbridgeWixel(Context context) {
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
         String collection_method = prefs.getString("dex_collection_method", "BluetoothWixel");
@@ -38,6 +40,8 @@ public class CollectionServiceStarter {
         }
         return false;
     }
+    public static boolean isDexbridgeWixel(String collection_method) { return collection_method.equals("DexbridgeWixel"); }
+
     public static boolean isBTShare(Context context) {
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
         String collection_method = prefs.getString("dex_collection_method", "BluetoothWixel");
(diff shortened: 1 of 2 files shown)


tzachi-dar:
@StephenBlackWasAlreadyTaken Seems to me that it is not enough. We have to pass the collection_method to all the 4 functions isBTWixel(), isDexbridgeWixel() and so on. Since java does not have a default parameter and this functions are called a lot of time I believe that another set of four functions should be created. For example:
 public static boolean isBTWixel(Context context) {
    SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
    String collection_method = prefs.getString("dex_collection_method", "BluetoothWixel");
    return isBTWixel(context, collection_method)
}

public static boolean isBTWixel(String collection_method) {
if(collection_method.compareTo("BluetoothWixel") == 0) {
        return true;
    }
    return false;
}

StephenBlackWasAlreadyTaken:
Ah good call, will update!

AdrianLxM:
@tzachi-dar's version jlooks good to me to pass the updated value (as in the settings there will be the value only AFTER onPreferenceChange retruns true.)

Just instead of
`public static boolean isBTWixel(String collection_method) {
if(collection_method.compareTo("BluetoothWixel") == 0) {
return true;
}
return false;
}`

I would write

`public static boolean isBTWixel(String collection_method) {
return collection_method.equals("BluetoothWixel");
}`

StephenBlackWasAlreadyTaken:
@AdrianLxM and @tzachi-dar can you check it out once more

tzachi-dar:
Well this should work.
I would have written it a little differently, but this is only an issue of style.

Tzachi

AdrianLxM:
Looking good to me :)

I just wonder why it worked before switching between bridge and BT-wixel... must have been the failover timer?

tzachi-dar:
two reasons why we never saw it before:
1) on start-sensor the services are also restarted.
2) bridge and BT-wixel are using the same service. they change things in the code, but they call isBT many times and get the correct result.

I only found it since I wanted to stop receiving packets and moved the input method to find that I still receive them.

jstevensog:
I probably should have made a separate service for xBridge.  I didn't at
the time, as essentially wixel-xDrip and xBridge did the same thing, but
using different protocols.  Can always revisit down the track, as it is
less likely to cause issues as the xBridge protocol evolves.

On Tue, Aug 4, 2015 at 3:47 PM, tzachi-dar notifications@github.com wrote:

> two reasons why we never saw it before:
> 1) on start-sensor the services are also restarted.
> 2) bridge and BT-wixel are using the same service. they change things in
> the code, but they call isBT many times and get the correct result.
> 
> I only found it since I wanted to stop receiving packets and moved the
> input method to find that I still receive them.
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/121#issuecomment-127486117
> .

## 

John Stevens

PR #125, opened 2015-08-05
Mongo update

AdrianLxM:
Update mongo-java-driver to 3.0.3 (latest version) that should work with the upcoming upgrade of the free sandbox databases to version 3.0.

It seems to work on my current sandbox database (version 2.6) as for the direct upload as for the data that got queued while no data connection was possible.
-  I tested with the chrome-plugin as viewer, not nightscout/azure -> please test if you have nightscout/azure
-  Is any more update necessary to handle the actual update of the db to 3.0?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -97,7 +97,7 @@ dependencies {
     compile 'com.android.support:appcompat-v7:22.1.1'
     compile 'com.squareup.okhttp:okhttp:2.2.0'
     compile 'com.google.code.gson:gson:2.3'
-    compile 'org.mongodb:mo'
+    compile 'org.mongodb:mongo-java-driver:3.0.3'
     compile 'com.squareup.retrofit:retrofit:1.9.0'
     compile 'com.getpebble:pebblekit:3.0.0'
     compile 'io.reactivex:rxjava:1.0.0'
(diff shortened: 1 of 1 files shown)


StephenBlackWasAlreadyTaken:
Nice, we will need to keep an eye on memory leaks but this should do the trick!

StephenBlackWasAlreadyTaken:
Merge when ready!

AdrianLxM:
@StephenBlackWasAlreadyTaken: How soon do we need it? I'd like to test it for at least 3 consecutive days first.

StephenBlackWasAlreadyTaken:
Well, I would like to get the next beta update out today, so we can wait until the end of the beta and put this in before we cut it to the main version, but it might be better to have many people testing it while its beta status. I could go either way

StephenBlackWasAlreadyTaken:
Or
we can just have a third beta update to be safe (would just include this and any other bugs that may pop up), that might be our best course of action

AdrianLxM:
I'm definitely in favour of at least the option of a third beta update.
But with the old db-version it looks good so far - so if you want you can merge for the current beta 2.2 and when the actual test with db-version 3.0 should fail we'd have to react anyways.

tzachi-dar:
I'll soon update one or two dexterity versions to use the latest drivers
and also an xdrip.
the xdrip will be using the wifi version that goes to the DB much more
times so I guess that in 3 days I can fake the load of 30 days of the
normal xdrip work.

This should give us a better understanding of how things are going
relatively fast.

By the way I also think that if one opens a new mogno account he will get
the 3.00 version right now. I might try that if I have time.

Thanks
Tzachi

AdrianLxM:
worked for me until now. @tzachi-dar did a test with much heavier load -> merging

PR #126, opened 2015-08-07
Add rendering of calibration values on graph.

JacoCronje:
This renders the calibration points on the graph. They are slightly larger points than the cgm points, white outside with red dot inside. It is useful to see when and where the calibration happened during the last 24 hours.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
@@ -23,6 +23,7 @@ import com.google.gson.GsonBuilder;
 import com.google.gson.annotations.Expose;
 import com.google.gson.internal.bind.DateTypeAdapter;
 
+import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


AdrianLxM:
I like it. It looks good on my phone. The code looks straight forward so I don't see any possible problems. :)

The calibration points overlap/cover the "bg readings" (sensor reading). I've tried to change the order to make the sensor reading cover the calibration. It works as calibration points are bigger but it doesn't look that good. :(

Good to go from my side.

AdrianLxM:
Now I've got it like this:
- Red dot is "on top"
- White circle can be overpainted by bg readings

Screenshot: image with calibration near bg vlaue
What do you think @JacoCronje ?

---

``` java
    public List<Line> defaultLines() {
        addBgReadingValues();
        List<Line> lines = new ArrayList<Line>();
        Line[] calib = calibrationValuesLine();
        lines.add(calib[0]); // white circle of calib in background
        lines.add(minShowLine());
        lines.add(maxShowLine());
        lines.add(highLine());
        lines.add(lowLine());
        lines.add(inRangeValuesLine());
        lines.add(lowValuesLine());
        lines.add(highValuesLine());
        lines.add(rawInterpretedLine());
        lines.add(calib[1]); // red dot of calib in foreground
        return lines;
    }
``

After almost one day of testing I can say: This is really really useful! Thanks a lot!
```

JacoCronje:
The white circle behind the rest can work. I don't mind having the BG value on top of the CGM value because the BG value is suppose to be more accurate and superior.
I'm happy if I can to see the calibration BG values on the graph, don't mind if it overlaps a CGM value.

But you do have a valid point. I just had a calibration 5-10 minutes ago where I first thought I missed a value, but looking closely I saw a few pixels of blue hiding behind the white.

StephenBlackWasAlreadyTaken:
Nice, Mind updating the pr with @AdrianLxM suggestions!? Awesome stuff

JacoCronje:
Code updated with white circle drawn in background

StephenBlackWasAlreadyTaken:
Thanks!!

PR #127, opened 2015-08-11
Fix sensor duration shown in system status. 

JacoCronje:
The duration(int) was overflowing when the sensor is more than 24 days old.

Before fix: Duration incorrectly shown as minus 24 days.
before_fix

After fix: Duration correctly shown as 25 days 3 hours.
after_fix


AdrianLxM:
Nice catch, thanks! 

As to how I intended to solve it, I was missing brackets to just cast the result to int (truncateing) that is now done automatically as all operands of the division are (long) integers.

**I support this pr.**

StephenBlackWasAlreadyTaken:
Sweet! Looks good, mind merging it in Adrian, I hate the mobile github view

AdrianLxM:
merged, thanks!

PR #130, opened 2015-08-16
Transmitter status in status

AdrianLxM:
... please comment

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/SystemStatus.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/SystemStatus.java
@@ -120,18 +120,19 @@ public class SystemStatus extends ActivityWithMenu {
 
         if (td== null || td.sensor_battery_level == 0){
             transmitter_status_view.setText("not available");
+        } else if((System.currentTimeMillis() - td.timestamp) > 1000*60*60*24){
+            transmitter_status_view.setText("no data in 24 hours");
         } else {
             transmitter_status_view.setText("" + td.sensor_battery_level);
 
             if (td.sensor_battery_level <= 207) {
-                transmitter_status_view.append(" (empty) ");
-            } else if (td.sensor_battery_level <= 213) {
-                transmitter_status_view.append(" (ok - on the low side) ");
+                transmitter_status_view.append(" - empty");
+            } else if (td.sensor_battery_level <= 210) {
+                transmitter_status_view.append(" - ok - rather low");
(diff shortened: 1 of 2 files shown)


ghost:
Would be very awesome a integration of the transmitter status!

Cheers,
Matthias

tzachi-dar:
We are missing here a lot of data but here is what I think:

Basing the data on one packet is problematic, it will not be consistent.
213 should be a good value.
My sensor was working with 212 for more than 2 month, so it might be a good value as well...

As for a real bug: you check for ==213 and < 212 which means you are missing the 212...

Another thing is that this requires the user to do something to look at it. I think that we should be proactive in this case. Maybe we should call something like: RisingAlert().
Please note that although this looks simple it also needs changes in UserNotification.create (and maybe also UI).
In any case it's ok for me even without an alert.

AdrianLxM:
@tzachi-dar, good catch!

I changed "empty" to 207 (according to @jstevensog / Adrien LeCroy) this is the value the official receiver shows it.

Another question is where to put the "on the low side" interpretation. How long was it running after it reached 111 then?

Yes, I've seen in the db: The number may jump back and forth between two neighbouring values. For the status this is fine imho, but for an alarm we'd have to think about something. And how to trigger the alarm would certainly be your domain of expertise, @tzachi-dar :)
For now the system status is at least something  - and I hope to gather some more knowledge/experience with the number, once people can actually look at it if they want.

tzachi-dar:
Where is the new code?
I'll check how long it was running from 111 once I reach home.

Thanks
Tzachi

AdrianLxM:
> Where is the new code?
> Here finally :)
> My still open terminal shows me that I did the push "successfully" about 6 hours ago. Now it is really online.
> 
> I'll check how long it was running from 111 once I reach home.
> That would be nice. What do you think would be a good timeframe?
> I'd like to have about 1 month - and if that is not possible, the value where it really starts dropping.

tzachi-dar:
OK, so this looks good. The only thing to do once I'm home is to look at the numbers and see what number will give the user a month. Although, that will probably change once we have more data.

AdrianLxM:
Got OK from @tzachi-dar and  @jstevensog on gitter... merging.

StephenBlackWasAlreadyTaken:
Nice, thanks for this!

ghost:
Great job, awesome ...just installed the apk

screenshot_2015-08-18-17-20-19

Thanks guys!

Cheers,
Matthias

PR #131, opened 2015-08-18
Transmitter status home

AdrianLxM:
- Added Transmitter Battery Status to Homescreen
- It will be in the background - chart is drawn on top (-> should not affect visibility much)
- Sensor will now have lowest transmitter battery level reported so far for this sensor period.
- Moved constants for thresholds to class Constants.java (avoid redundancy/inconsistency)

This is how it looks like:

low
empty

Should we add "experimental interpretation" to the "rather low"-status like in the SystemStatus?
Feedback highly appreciated! :)

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -5,13 +5,19 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.content.SharedPreferences;
+import android.graphics.Canvas;
 import android.graphics.Color;
+import android.graphics.ColorFilter;
 import android.graphics.Paint;
+import android.graphics.drawable.Drawable;
+import android.graphics.drawable.ShapeDrawable;
+import android.graphics.drawable.shapes.Shape;
 import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.support.v4.widget.DrawerLayout;
+import android.util.DisplayMetrics;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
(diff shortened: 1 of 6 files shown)


ghost:
Good idea, I would prefer an alarm/notification in terms of transmitter battery status.

Cheers,
Matthias

AdrianLxM:
@mgranberry @StephenBlackWasAlreadyTaken: Changed status message to "low" and "VERY LOW".

@ceben80, At least the "low" status should be shown when there is still about one month of battery. This is why we agreed not to have an alarm for that - that could ring during the night or at a meeting, when it is not actually urgent to react.
An alarm for the "Very low"/empty state would be a different pr. The future will tell if it is even necessary as users are hopefully advised early enough now.

AdrianLxM:
Tested when setting new sensor and the battery_level gets updated correctly.

StephenBlackWasAlreadyTaken:
This looks good to me, do we plan on ever using the background to display other alerts?

AdrianLxM:
At the moment I cannot think of other not rapidly changing notifications that should be shown but are not urgent enough to rise a Notification/Alarm.

jstevensog:
Would be good if these messages also appeared in widget and permanent notification, so they are always visible.
Just a thought.
Cheers

---- AdrianLxM wrote ----

StephenBlackWasAlreadyTaken:
Merge at will!

StephenBlackWasAlreadyTaken:
(sorry it took me a while to get around to changing my wixel to send low values for testing)

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -149,6 +154,40 @@ public void setupCharts() {
 
         chart.setZoomType(ZoomType.HORIZONTAL);
 
+        //Transmitter Battery Level
+        final Sensor sensor = Sensor.currentSensor();
+        if (sensor != null && sensor.latest_battery_level != 0 && sensor.latest_battery_level <= Constants.TRANSMITTER_BATTERY_LOW) {
```
StephenBlackWasAlreadyTaken, line 159: actually should we put in a check here to never show anything if they are using dexcom share

AdrianLxM:
@StephenBlackWasAlreadyTaken, I thought, that with checking for null and if the battery level is not 0, not showing anything for share users is done automatically.
I cannot test this though.

Did you take back the "Merge at will!"?

StephenBlackWasAlreadyTaken:
I suppose just checking that its not zero should be fine!

AdrianLxM:
I totally missed your last comment... merged

PR #133, opened 2015-08-23
Fixes for xBridge Prompt

AdrianLxM:
- Remove barcode scanner for share from xBridge prompt
- Make txid uppercase

Please comment.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/BluetoothScan.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/BluetoothScan.java
@@ -16,6 +16,7 @@ import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.preference.PreferenceManager;
+import android.text.InputFilter;
 import android.text.TextUtils;
 import android.util.Log;
 import android.view.LayoutInflater;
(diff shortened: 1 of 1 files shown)


StephenBlackWasAlreadyTaken:
We dont want to allow scanning?

AdrianLxM:
As far as I see it, there is no scanning for xBridge.
There wasn't even an onClickListener registered.

Also as far as I remember (not having Share), the barcode on the Share receiver is the serial number, not the transmitterID. So scanning would just be for Share, not for xBridge.

StephenBlackWasAlreadyTaken:
I believe @mgranberry added it for scanning the transmitter box

AdrianLxM:
Hm, I thought this was just reuse of the code/layout. The button was not active so for beta4 or release something should be changed?
What was your intention @mgranberry?

mgranberry:
Yes.  It was for scanning the transmitter box, but the serial number format was complex enough that I didn't want to do anything with it until I was able to look at a non-US box.  Removing it is probably the correct thing to do.

AdrianLxM:
Except from making it all caps on entering the code to avoid wrong user input, the button is just hidden with `dialog.findViewById(R.id.scannerButton).setVisibility(View.GONE);` but still present in the layout.

I can send you, @mgranberry, the photo of a German box tomorrow, if you like.

AdrianLxM:
I'll merge this - it just hides the button- easily undoable once it actually scans something.
Making it all caps I think should be in.

PR #135, opened 2015-08-26
Csv export

AdrianLxM:
This exports a CSV with columns date, time and cgms-reading in a format that SiDiary can import.

SiDiary is a bg diary that can import from or directly read almost all bg meters, pumps and CGMS on the market. Now also xDrip :)

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -13,6 +13,7 @@
     <string name="menu_stop">Stop</string>
     <string name="menu_refresh">Refresh</string>
     <string name="menu_export_database">Export database</string>
+    <string name="menu_export_csv">Export CSV</string>
     <string name="error_bluetooth_not_supported">Error: Bluetooth not supported by this device</string>
     <string name="unknown_device">Unrecognized Device</string>
     <string name="connecting_to_device">Connecting to device</string>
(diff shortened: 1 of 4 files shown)


StephenBlackWasAlreadyTaken:
Looks good to me, Havnt been able to test it yet, are there other csv formats we may want to support down the line?

AdrianLxM:
I've asked on gitter. So far there were none.
But following the same pattern it will be very fast to add different formats:
Just modify the header, add the fields needed to the query string and print them to the  output stream.

app/src/main/res/values/strings.xml
```diff
@@ -13,6 +13,7 @@
     <string name="menu_stop">Stop</string>
     <string name="menu_refresh">Refresh</string>
     <string name="menu_export_database">Export database</string>
+    <string name="menu_export_csv">Export CSV</string>
```
StephenBlackWasAlreadyTaken: Mind updating the text here to `Export CSV (SiDiary format)` otherwise it may confuse people why its so sparse on information

AdrianLxM:
added "(SiDiary format)". Thanks, this was a good idea.

StephenBlackWasAlreadyTaken:
Awesome thanks, Merge at will!

AdrianLxM:
Thank you!

ghost:
brilliant feature this csv export, thanks very much, great job

PR #142, opened 2015-09-03
start service on 4.3 also

skjelland:


tzachi-dar:
"Officialy" we indeed work on 4.3.

PR #144, opened 2015-09-07
text to speech

AdrianLxM:
If text to speech is installed on the system, it will read new values out loud if activated in settings.

Pitfalls for testing:
As the data usually is stored on the sc-card, it might not work when still connected via USB.
First reading might be a few seconds late as the text-to-speech system has to be initialized.

And please forgive me the "speach"  :scream:

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
@@ -169,7 +169,7 @@ public class Notifications extends IntentService {
                 AlertPlayer.getPlayer().ClockTick(context, trendingToAlertEnd, EditAlertActivity.unitsConvert2Disp(doMgdl, bgReading.calculated_value));
                 return;
             }
-            // Currently the ui blocks having two alerts with the same alert value.
+           // Currently the ui blocks having two alerts with the same alert value.
 
             boolean alertSnoozeOver = ActiveBgAlert.alertSnoozeOver();
             if (alertSnoozeOver) {
@@ -212,11 +212,11 @@ public class Notifications extends IntentService {
 
     boolean trendingToAlertEnd(Context context, Boolean newAlert, AlertType Alert) {
         if(newAlert && !smart_alerting) {
-            //  User does not want smart alerting at all.
+        //  User does not want smart alerting at all.
             return false;
         }
         if((!newAlert) && (!smart_snoozing)) {
-            //  User does not want smart snoozing at all.
+        //  User does not want smart snoozing at all.
             return false;
         }
         return BgReading.trendingToAlertEnd(context, Alert.above);
@@ -497,7 +497,7 @@ public class Notifications extends IntentService {
 
     public static void RiseDropAlert(Context context, boolean on, String type, String message, int notificatioId) {
         if(on) {
-            // This alerts will only happen once. Want to have maxint, but not create overflow.
+         // This alerts will only happen once. Want to have maxint, but not create overflow.
             OtherAlert(context, type, message, notificatioId, Integer.MAX_VALUE / 100000);
         } else {
             NotificationManager mNotifyMgr = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
@@ -271,7 +271,7 @@ public class Notifications extends IntentService {
 
         //Text to speech
         Log.d("BgToSpeech", "gonna call speak");
-        BgToSpeech.getSingleton(context).speak(bgReading.calculated_value);
+        BgToSpeech.getSingleton(context).speak(bgReading.calculated_value, bgReading.timestamp);
     }
 
     private void  ArmTimer() {
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
@@ -233,6 +233,15 @@ public class Notifications extends IntentService {
         if (bg_ongoing && (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN)) {
             bgOngoingNotification(bgGraphBuilder);
         }
+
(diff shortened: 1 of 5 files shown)


ghost:
An absolutely amazing feature. It works without any problems.
Especially in situations where you can not carry the Smart Watch, perhaps in dirty gardening...fantastic!

Matthias

AdrianLxM:
Added reinitialization as connection to TTS might get lost after some time in the background.
Hope this helps.
@ceben80, how's your test going? Would you mind updating to the latest version? Danke!

ghost:
Good morning!

So far, it works fine on my device...but of course I have been updating to the latest version 5 min. ago which is running now.

Cheers,
Matthias

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
```diff
@@ -232,6 +233,15 @@ public void notificationSetter(Context context) {
...
             bgOngoingNotification(bgGraphBuilder);
         }
+
+        List<BgReading> bgReadings = BgReading.latest(3);
+        if(bgReadings == null || bgReadings.size() < 1) { return; }
+        BgReading bgReading = bgReadings.get(0);
+
+        //Text to speech
+        Log.d("BgToSpeech", "gonna call speak");
+        BgToSpeech.getSingleton(context).speak(bgReading.calculated_value, bgReading.timestamp);
```
StephenBlackWasAlreadyTaken: Any reason this is here in the notifications and not in the BgSendQueue (where I figure all new values are going to be pushed through)
This however may get called at unexpected times, no?

app/src/main/java/com/eveningoutpost/dexdrip/utils/BgToSpeech.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    public void speak(final double value, long timestamp){
+        final SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+
+        if (! prefs.getBoolean("bg_to_speech", false)){
+            return;
+        }
+
+        if(this.timestamp == timestamp){
```
StephenBlackWasAlreadyTaken, line 53: ah is this how you guarded against it being called multiple times in a row if multiple things kicked off the service?

app/src/main/java/com/eveningoutpost/dexdrip/utils/BgToSpeech.java
```diff
@@ -0,0 +1,139 @@
...
+                text =  df.format(value);
+            } else {
+                df.setMaximumFractionDigits(1);
+                df.setMinimumFractionDigits(1);
+                text =  df.format(value* Constants.MGDL_TO_MMOLL);
+            }
+        } else if (value > 12) {
+            text =  "low";
+        } else {
+            text = "no value";
```
StephenBlackWasAlreadyTaken, line 133: do we want it to say no value on error codes?

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -59,6 +59,9 @@ private void refreshFragments() {
 
     @Override
     public void onActivityResult(int requestCode, int resultCode, Intent data) {
+
+
+
```
StephenBlackWasAlreadyTaken: can we clean up all this extra whitespace

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -127,9 +130,9 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-
+        preferenceFragment = new AllPrefsFragment();
```
StephenBlackWasAlreadyTaken, line 130: Is this change related? it dosnt seem like we use the preferenceFragment again unless Im missing it

AdrianLxM:
I put it in Notifications as I searched for where the ongoing notification was created and I saw it conceptually as a notification. Should I change this?

The preferenceFragment is a leftover from calling "findPreference" on that fragment. As we have a reference variable for it and it is set in "refreshFragments()" I would keep it for consistency or fully delete it. Sorry for the whitespaces. I'll clean them up.

About the "no value" - I'm not sure about this. What is better? Nothing at all? Or "Error"? (Error exists in many languages).

AdrianLxM:
cleaned up the whitespaces.
And yes, I use the timestamp to guard against it being called multiple times.

StephenBlackWasAlreadyTaken:
I would prefer it get called from the send queue, that way all the "Hey theres a new value, kick off these things" gets called from the same place. I would also wonder how this will work for a dexcom share user (they will pull down about 20 or so values quickly if its been disconnected for a bit and obviously we dont want it reading all of them as quickly as its processing them)

As for the rest, Im alright with keeping it as you have it, maybe instead of no value, error value?

AdrianLxM:
I put it in BgSendQueue. Works fine so far with xDrip-wixel.

Any ideas on the share input? I could not start (queue) new text-to-speech output, if it is still speaking. This would at least reduce the amount of values spoken. But this would mean, that the first value is spoken and not the current one.

I would rather not do the distinction in the BgReading.create()s.

StephenBlackWasAlreadyTaken:
Maybe have it only read it if the timestamp of the bg value is from within the last 4(ish) minutes?

AdrianLxM:
Yea right. When I think "that's too complicated to do", you have to come up with "the obvious and simple" solution. ;)
Thanks. I'll implement it tomorrow.

StephenBlackWasAlreadyTaken:
haha, Thanks for this! I look forward to getting it merged in!

AdrianLxM:
I think it will need some more rewriting and testing before merging. :/

On my phone with low memory, after not using tts for a while, it cannot reconnect to the tts engine. I think we have to initiate it from the collection services and properly shut it down again on onDestroy().

AdrianLxM:
I now initiate the tts system in the collection services - and disconnect it again on onDestroy().
This should lead to more stability.
New branch for that: https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/tree/tts2

AdrianLxM:
I'm closing this in favour of the new branch that runs stable on my phone: https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/146

PR #146, opened 2015-09-10
Text-to-speech 2

AdrianLxM:
With proper initialization in the DexCollectionService and release of the resource in onDestroy(), it runs stable on my phone.

@tzachi-dar and @jamorham (you are the only two I know that use Wifi-Wixel): Does this work for you, as WifiWixel is a thread not a service?

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
@@ -44,6 +44,7 @@ import com.eveningoutpost.dexdrip.Sensor;
 import com.eveningoutpost.dexdrip.UtilityModels.CollectionServiceStarter;
 import com.eveningoutpost.dexdrip.UtilityModels.ForegroundServiceStarter;
 import com.eveningoutpost.dexdrip.UtilityModels.HM10Attributes;
+import com.eveningoutpost.dexdrip.utils.BgToSpeech;
 
 import java.nio.ByteBuffer;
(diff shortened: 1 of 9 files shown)


tzachi-dar:
Did not have the time to test it yesterday, will test it today.

Thanks
Tzachi

tzachi-dar:
Tested it with wifi wixel. Works great.

StephenBlackWasAlreadyTaken:
Seems to work for me just fine, I didnt test mmol but I assume one of you did no?
If so, go ahead and merge at will!

AdrianLxM:
Thanks for testing! I was stable the second night in a row and even worked fine on my ZGPAX S8 watch.
I tested it with mmol/l + German locale as well.

tzachi-dar:
By the way,
I have also noticed that if I want to play american English, than this
seems to be already installed than I did not really have a way to set it,
and it did not work. Only when I choose UK English, I was given a chance to
install it and then it worked.

Not sure if this is related to us or an android problem, but just wanted
you to know.

Thanks
Tzachi

AdrianLxM:
Thanks for reporting. But I guess this is quirk of the Android tts system.

How I programmed it, you cannot set the language in xDrip.
It will try the locale set in the Android operating system and if that does not work, it will try generic English (should be  Australian, British,  Indian, US ... in theory what ever you've installed. But maybe it chooses british on your system for some reason?
I don't know how Android prioritizes one English accent over the other. I leave this up to the standard behaviour of Android. If you want a specific English, I guess you could set the locale of the system to that one.

PR #149, opened 2015-09-14
Age adjust filtered

AdrianLxM:
I hope this solves the problem with NS showing "raw" data with xDrip (maybe not yet xBridge)
(Hope as in: I cannot test it, but it seems very very likely to me that this is the problem.)

Thanks @jstevensog for taking over from here :)

Nightscout adjusts the raw datum with a ratio it calculates from the sgv and the filtered datum (the datum the sgv usually is calculated with). As xDrip calculates it with an age adjusted value but still uploads a non age adjusted value as filtered datum, the ration may become not 1 where it should be.
https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/plugins/rawbg.js#L59

xBridge:
As the filtered value for xBridge is acutally set and not a duplicate of raw_data, I just can think of two solutions:

a) Upload "0" as filtered - then no factor will be calculated in NS
b) Upload "age_adjusted_raw" as "filtered".

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -1000,6 +1000,16 @@ public class BgReading extends Model {
         return age_adjusted_raw_value;
     }
 
+    public double ageAdjustedFiltered(){
+        double usedRaw = usedRaw();
+        if(usedRaw == raw_data || raw_data == 0d){
+            return filtered_data;
+        } else {
+            // adjust the filtered_data with the same factor as the age adjusted raw value
+            return filtered_data * (usedRaw/raw_data);
+        }
+    }
+
     // the input of this function is a string. each char can be g(=good) or b(=bad) or s(=skip, point unmissed).
     static List<BgReading> createlatestTest(String input, Long now) {
         Random randomGenerator = new Random();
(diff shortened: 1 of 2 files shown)


StephenBlackWasAlreadyTaken:
People seem to be confused about 'raw' data as it shows in nightscout and want to see it when using xdrip and xbridge, many people are happy to have it in xbridge and I cant seem to explain to them enough how its not going to be helpful. Do we expect a lot of upset people who 'loose' their raw data if we move this in?

AdrianLxM:
@StephenBlackWasAlreadyTaken:
How I understand this: (https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/plugins/rawbg.js#L59)
With xDrip the raw data in the first two days will be gone. After that there should not have been raw data anyways.
I'm not totally sure about that: With xBridge that uploads the filtered data but still calculates the sgv from the unfiltered one, there will still be "raw" data like standard nightscout shows when dexcom decides to calculate the sgv is calculated from the unfiltered datum.
How I interpret the code in NS, it squares the factor between filtered and unfiltered to show the raw data.

sgv = unfiltered // in this case
ratio = filtered / sgv = filtered/unfiltered
raw = unfiltered/ratio = unfiltered^2 / filtered

from our chat yesterday, @jstevensog seems to prefer that behaviour. It would basically mean we have to tell the people: "xBridge values are raw but the raw values are a trend prediction"?

About taking away raw: It wasn't really there with xDrip in the first place and with xBridge it should behave the same - with the exception of the first 2 days.

Disclaimer: I cannot test this as I still have no NS setup. Sorry If I'm mistaken. I just thought I found the reason... I just leave this here and let you guys handle it (and if it is still an issue when I get finally find time to setup NS, I'll join again)

StephenBlackWasAlreadyTaken:
ah, your explanation makes a lot of sense. Im fine with this getting merged in. Im not someone who turns the raw on other than for testing purposes but I dont see anything wrong with this change

AdrianLxM:
I'll merge this then.

PR #152, opened 2015-09-18
Using Colour.TRANSPARENT in v.setDrawingCacheBackgroundColor 

jstevensog:
instead of 0.

Improves readability.  Only a minor change.


AdrianLxM:
Color.TRANSPARENT == 0

I like this. Improves readability and won't have any problems if the framework changs the colour constants in future versions (not very likely, but still).

StephenBlackWasAlreadyTaken:
Works for me! Readability is important, Merge at will!

PR #153, opened 2015-09-18
Altered RetryTime time for xBridge.

jstevensog:
I found the 65 second retry time was too long for xBridge at times.  Reducing it to 25 seconds has given a good boost to performance.  I believe that Android often gets too busy to allow the retry timer to be reliable, and sometimes xDrip does not re-establish the connection to a newly woken xBridge in time to get the data packet.  This change has proven this thought to be somewhat correct.

I decided not to make it shorter than this to keep the logs from filling up with messages about it.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
@@ -149,7 +149,12 @@ public class DexCollectionService extends Service {
 
     public void setRetryTimer() {
         if (CollectionServiceStarter.isBTWixel(getApplicationContext()) || CollectionServiceStarter.isDexbridgeWixel(getApplicationContext())) {
-            long retry_in = (1000 * 65);
+            long retry_in;
+            if(CollectionServiceStarter.isDexbridgeWixel(getApplicationContext())) {
+                retry_in = (1000 * 25);
+            } else {
+                retry_in = (1000*65);
+            }
             Log.d(TAG, "setRetryTimer: Restarting in: " + (retry_in/1000)  + " seconds");
             Calendar calendar = Calendar.getInstance();
             AlarmManager alarm = (AlarmManager) getSystemService(ALARM_SERVICE);


StephenBlackWasAlreadyTaken:
Isnt the retry timer just being called if A) the bt device is not connected, B) the service gets killed or C) the bt device disconnects?

I feel like relying on 25 second intervals to recheck if the device is back on seems a bit odd, Since I assume the main case here is because the xBridge disconnects itself for a set number of seconds can we instead have some new logic just in the device disconnecting use case that sets a very specific timer (ie if we know its going to be disconnected for 4 minutes right after it receives a new reading, maybe set it for for minutes at that time?)

jstevensog:
The problem with that is, say, if the xBridge is sleeping for 260 seconds,
and we set the retry timer to that, and for some reason xBridge goes back
to sleep in that time (only stays awake for 120 seconds after a packet is
captured), there is a potential gap in time between packet capture by
xBridge and receiving it by xDrip of up to 260 seconds.
As well, if it is totally out of sync for any reason, it xBridge will go
back to sleep waiting for the reconnection, and the packet will never
arrive at xDrip.

I actually see both these scenarios with the 65 second setting.  On
occasion, if android is busy doing other things(downloading app updates is
a good one), it will take way longer than 65 seconds to reconnect, more
than 120 seconds, and by that time xBridge has gone back to sleep and not
sent a packet.  Also, the trend is all over the place in time intervals.

If we are going to stick with one value, I would recommend 25 seconds, or
even lower.  The trend is more evenly time spaced, and the chances of
xBridge going to sleep waiting for android to get around to servicing the
GATT are much lower.

GATT appears to have much lower priority to things like WiFi and phone
functions.  Hence why it doesn't pick up that the BLE device is ready for
reconnection until the retryTimer expires and it gets a kick.  If we could
make android prioritise the BLE, or at least rely on some sort of
determinism, we wouldn't need it.  IMHO, the Bluetooth reconnection should
be MUCH faster.  An example would be a headset going out of or coming back
into range.  It should be handled almost instantaneously.  I am guessing
that the Pebble (Time) apps have a retry like 5-10 seconds, and even they
sometimes struggle to reconnect.

On Fri, Sep 18, 2015 at 11:38 PM, Stephen Black notifications@github.com
wrote:

> Isnt the retry timer just being called if A) the bt device is not
> connected, B) the service gets killed or C) the bt device disconnects?
> 
> I feel like relying on 25 second intervals to recheck if the device is
> back on seems a bit odd, Since I assume the main case here is because the
> xBridge disconnects itself for a set number of seconds can we instead have
> some new logic just in the device disconnecting use case that sets a very
> specific timer (ie if we know its going to be disconnected for 4 minutes
> right after it receives a new reading, maybe set it for for minutes at that
> time?)
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/153#issuecomment-141454075
> .

## 

John Stevens

StephenBlackWasAlreadyTaken:
How far ahead of a packet being received by the wixel does it wake up the ble module?

jstevensog:
About 40 seconds. This could be reduced if the retry timer was lower.  30 seconds if retry is 25, as we would really like the connection established when it receives a packet.  
10 seconds with a retry of 5, etc.

I just wish the Android BLE was a bit smarter.  Really, it would be best if the GATT layer took care of this.

---- Stephen Black wrote ----

jstevensog:
Personal opinion is that for a medical device, reconnection should occur
almost immediately.  Regardless of if the connection is the result of a
device sleeping, going our of range, or losing power.  I would certainly
advocate strongly for a retry timer of 5 seconds to meet that.  But I am
unsure of what impact it would have on the overall performance of the GATT
and collection service.

Around every 3-5 days I have to manually restart the collection service, as
xBridge is sending the packets, but the service is completely ignoring
them, or so it appears.  As it is rare, and I am keen to immediately
restart the service (usually away from all my diagnostic equipment when it
happens) I am yet to find out precisely what is going on.  But it would
appear the 6 minute Failover timer doesn't initiate.  Likely I just haven't
gone deeply through the code to figure out what triggers it.
Cheers

On Sat, Sep 19, 2015 at 8:18 AM, John Stevens jstevensog@gmail.com wrote:

> About 40 seconds. This could be reduced if the retry timer was lower.  30
> seconds if retry is 25, as we would really like the connection established
> when it receives a packet.
> 10 seconds with a retry of 5, etc.
> 
> I just wish the Android BLE was a bit smarter.  Really, it would be best
> if the GATT layer took care of this.
> 
> About 40 seconds. This could be reduced if the retry timer was lower.  30
> seconds if retry is 25, as we would really like the connection established
> when it receives a packet.
> 10 seconds with a retry of 5, etc.
> 
> I just wish the Android BLE was a bit smarter.  Really, it would be best
> if the GATT layer took care of this.
> 
> ---- Stephen Black wrote ----
> 
> How far ahead of a packet being received by the wixel does it wake up the
> ble module?
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/153#issuecomment-141579886
> .

## 

John Stevens

AdrianLxM:
Depending on the chipset and BT-stack used, Android devices can take even more than 10 seconds to (re-)connect. I'm not sure if reconnects (calling attemptConnection...) every 5 seconds would be good as that would be a second call before most devices even had the chance to actually connect.

I'm hoping BLE will work better in newer Android versions.

StephenBlackWasAlreadyTaken:
Yeah, Im wondering how it handles the case where you begin a connection just as your retry timer gets called, I imagine you are going to start locking up the gatts in a bad way but Im not sure how big of a window that actually is, remember that each callback happens on a different thread and they dont really know about each other

tzachi-dar:
Seems to me that things are working quite well, most of the time (more than
a few days), but then having some trouble, So I wander if we should not
keep the behavior as it is right now, but try to detect this situations and
only than try harder the reconnect.
This can happen in the xDrip, reconnecting faster or on the xBridge staying
on for say 20 minutes to allow the sync again.

By the way, the packets the transmitter sends should be very 5 minutes from
each other, so if one remembers the time of one packet, he can know when
the next one will be, so he should know when the ble model is up. No need
to reconnect at other times.

Thanks
Tzachi

jstevensog:
What I have discovered, after using xDrip with xBridge for a while now, is there are 3 reasons a packet is dropped. 
1) wixel misses a packet.  (Out if range, or interference)
2) BLE doesn't reconnect within 2 minutes of a packet capture.  Rare, and reducing the retry timer to 25 seconds virtually eliminated this.
3) DexCollectionService appears to stop processing packets and requires a restart to fix.  BLE is established in this case, and I usually only know about it when I get the 16 minute "missed packets" notification. This one needs more diagnosis.

Reason 2 appears to occur when the phone processor us busy doing other things, and even though the retry timer has expired, DexCollectionService doesn't get processor time to do the retry.  So I tried reducing the timer delay to 25/seconds and it improved things greatly, proving the theory.

You can never see this running the debugger.  The debugger ensures the xDrip processes get cpu time.

For xBridge at least, this should be 25 seconds or below.  

Happy to hear about ideas to investigate 2&3, but a fix for 2, even interim, is available.
Cheers

---- tzachi-dar wrote ----

AdrianLxM:
From my point of view, 25 seconds for xBridge would be fine.
I just don't want to make this period that short that the actual connects might take longer.

StephenBlackWasAlreadyTaken:
thats my fear as well @AdrianLxM. @jstevensog  If you want to give it a go for the next few days and let us know if it hits that 'needing to be restared' more frequently with this change I am alright with it, but I definitely want to make sure its been vetted for a bit before merging it

jstevensog:
I've been running the 25 second retrybsunce Thursday (now Tuesday here) with absolutely no issues with reconnection delays.  If we want linger I'm happy to keep it running till the end if the week.  This has (I believe) almost entirely eliminated case 2.  Case 1, can't do much about, and case 3 will require different investigations to determine the cause.  But I think case 3 will come down to something in Android.

---- Stephen Black wrote ----

jstevensog:
Hi All,
It is now Friday, and I still have no issue with the 25 second retry time.  I would like to merge this in now.  If it is OK, please respond with LGTM.
I will be concentrating on the dumber watch face and looking into the DexCollectionService occasional "lock up" from now on.  Seems to be phone/Android dependent, as some see it far worse than I do, and have to reboot the phone to fix it, where I just have to restart the service.  BTW, those seeing this worse than I are running standard xDrip-E.
Cheers

StephenBlackWasAlreadyTaken:
@jstevensog still going strong? if so I think we are safe to merge

jstevensog:
Definitely going strong.  There is no impact of the 25second retry time, other than better packet capture.  Still have to figure out what is going on when the service stops accepting packets.  Not related to the retry time, and definitely not predictable.  Also, appears to depend on the phone model, so likely related to BLE hardware or driver. 
Taking a sensor break this week, so will work on it next week.
Cheets

---- Stephen Black wrote ----

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -149,7 +149,12 @@ public void listenForChangeInSettings() {
 
     public void setRetryTimer() {
         if (CollectionServiceStarter.isBTWixel(getApplicationContext()) || CollectionServiceStarter.isDexbridgeWixel(getApplicationContext())) {
-            long retry_in = (1000 * 65);
+            long retry_in;
+            if(CollectionServiceStarter.isDexbridgeWixel(getApplicationContext())) {
+                retry_in = (1000 * 25);
```
mgranberry, line 154: Codacy Issue found: [Useless parentheses.](https://www.codacy.com/app/matthias-granberry/xDrip-Experimental/file/1301958408/issues/source?bid=2735053&fileBranchId=2947766#l154)

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -149,7 +149,12 @@ public void listenForChangeInSettings() {
...
+            } else {
+                retry_in = (1000*65);
```
mgranberry, line 156: Codacy Issue found: [Useless parentheses.](https://www.codacy.com/app/matthias-granberry/xDrip-Experimental/file/1301958408/issues/source?bid=2735053&fileBranchId=2947766#l156)

jstevensog, line 156: Yeah, fair enough.  I do kinda like the readability of (1000 \* X) though.
There is a lot of this sort of thing throughout xDrip.  Normally I just put
the final number in and forget about the calculation, but I am pedantic
about not using CPU cycles for anything but what they are needed for ;)
Fix it up if you like.
Cheers

On Wed, Nov 11, 2015 at 6:40 AM, Matthias Granberry <
notifications@github.com> wrote:

> In
> app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/153#discussion_r44454252
> :
> 
> > @@ -149,7 +149,12 @@ public void listenForChangeInSettings() {
> > 
> > ```
> >  public void setRetryTimer() {
> >      if (CollectionServiceStarter.isBTWixel(getApplicationContext()) || CollectionServiceStarter.isDexbridgeWixel(getApplicationContext())) {
> > ```
> > -            long retry_in = (1000 \* 65);
> > -            long retry_in;
> > -            if(CollectionServiceStarter.isDexbridgeWixel(getApplicationContext())) {
> > -                retry_in = (1000 \* 25);
> > -            } else {
> > -                retry_in = (1000*65);
> 
> [image: Codacy]
> https://camo.githubusercontent.com/b1b1514cd8a7d7c9c9fa32f769284569ffee2882/68747470733a2f2f7777772e636f646163792e636f6d2f76657273696f6e65642f696d616765732f66617669636f6e2e706e67
> Issue found: Useless parentheses.
> https://www.codacy.com/app/matthias-granberry/xDrip-Experimental/file/1301958408/issues/source?bid=2735053&fileBranchId=2947766#l156
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/153/files#r44454252
> .

## 

John Stevens

jstevensog, line 156: But I would just get rid of the whole calculation if I was going to do
anything.
Cheers

On Wed, Nov 11, 2015 at 11:18 AM, John Stevens jstevensog@gmail.com wrote:

> Yeah, fair enough.  I do kinda like the readability of (1000 \* X) though.
> There is a lot of this sort of thing throughout xDrip.  Normally I just put
> the final number in and forget about the calculation, but I am pedantic
> about not using CPU cycles for anything but what they are needed for ;)
> Fix it up if you like.
> Cheers
> 
> On Wed, Nov 11, 2015 at 6:40 AM, Matthias Granberry <
> notifications@github.com> wrote:
> 
> > In
> > app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
> > https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/153#discussion_r44454252
> > :
> > 
> > > @@ -149,7 +149,12 @@ public void listenForChangeInSettings() {
> > > 
> > > ```
> > >  public void setRetryTimer() {
> > >      if (CollectionServiceStarter.isBTWixel(getApplicationContext()) || CollectionServiceStarter.isDexbridgeWixel(getApplicationContext())) {
> > > ```
> > > -            long retry_in = (1000 \* 65);
> > > -            long retry_in;
> > > -            if(CollectionServiceStarter.isDexbridgeWixel(getApplicationContext())) {
> > > -                retry_in = (1000 \* 25);
> > > -            } else {
> > > -                retry_in = (1000*65);
> > 
> > [image: Codacy]
> > https://camo.githubusercontent.com/b1b1514cd8a7d7c9c9fa32f769284569ffee2882/68747470733a2f2f7777772e636f646163792e636f6d2f76657273696f6e65642f696d616765732f66617669636f6e2e706e67
> > Issue found: Useless parentheses.
> > https://www.codacy.com/app/matthias-granberry/xDrip-Experimental/file/1301958408/issues/source?bid=2735053&fileBranchId=2947766#l156
> > 
> > —
> > Reply to this email directly or view it on GitHub
> > https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/153/files#r44454252
> > .
> 
> ## 
> 
> John Stevens
> "You are how you live, not what you have."

## 

John Stevens

jstevensog:
Closing.  This has been put into xBridgeImprovements which has been merged.

PR #157, opened 2015-09-24
Mongo REST upload

AdrianLxM:
TESTING - please do NOT merge

With the mongo 3 update we will lose the ability to directly upload to mongolab as the new java-driver does not run on Android.

As some people (including me) would still prefer direct Mongolab upload to the upload via Nightscout (not using NS, using Mongolab just to mirror data or to view on Garmin watch, ...), I wrote a first version of a REST-Uploader for MongoLab. The Rest-API upload is used by the chrome uploader as well and seems to be pretty stable.
We also may use this as a starting point to rewrite the NS-Rest-Upload, as they are pretty similar (code borrowed from current Rest-Upload).

So far it worked for me and takes 19930.0 ms to upload 32 records.

For testing:
- enable mongolab in settings (url does not matter)
- in class com.eveningoutpost.dexdrip.UtilityModels.MongoLabRest:
  - enter your database name as string constant DATABASE
  - enter your apiKey as string constant API_KEY (http://docs.mongolab.com/data-api/#authentication)

It would be great if you could test that! If it works, we can move the db-name and API-Secret to the settings and clean the old code from NightscoutUploader.
(@tzachi-dar - this is a newer version than I mailed you.)

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoLabRest.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoLabRest.java
@@ -78,7 +78,7 @@ public class MongoLabRest {
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(ctx);
         String dbName = prefs.getString("cloud_storage_mongodb_rest_database", "nightscout");
         String apiKey = prefs.getString("cloud_storage_mongodb_rest_key", "");
-        
+
         return new MongoLabRest(dbName, apiKey, ctx);
     }
 
@@ -129,7 +129,7 @@ public class MongoLabRest {
         if (!populateJasonMBG(json, cal)) {
             return false;
         }
-        
+
         String dsCollectionName = prefs.getString("cloud_storage_mongodb_collection", "entries");
         return sendToMongo(dsCollectionName, json);
     }
@@ -187,7 +187,7 @@ public class MongoLabRest {
             Log.e(TAG, "sendToMongo Exception: ", e);
             return false;
         }
-        
+
         return success;
     }
 
@@ -230,8 +230,11 @@ public class MongoLabRest {
 
 
     private boolean populateJasonDeviceStatus(JSONObject json, int batteryLevel) {
+        SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
+        format.setTimeZone(TimeZone.getDefault());
         try {
             json.put("uploaderBattery", batteryLevel);
+            json.put("created_at", format.format(System.currentTimeMillis()));
             return true;
         } catch (JSONException e) {
             return false;
@@ -261,31 +264,31 @@ public class MongoLabRest {
             return false;
         }
     }
-    
-    
-    
+
+
+
     public List<TransmitterRawData> fromJson(Context ctx, String data) {
-        
+
         Gson gson = new Gson();
         Type listType = new TypeToken<List<TransmitterRawData>>(){}.getType();
-        
-        List<TransmitterRawData> trdList = gson.fromJson(data, listType); 
-        
+
+        List<TransmitterRawData> trdList = gson.fromJson(data, listType);
+
         //Toast.makeText(ctx, "objects read " + asd.size(), Toast.LENGTH_LONG).show();
         Log.e(TAG,  "objects read " + trdList.size());
         return trdList;
-        
-        
+
+
     }
-    
-    
+
+
     // This function is based on    
     //http://www.javacodegeeks.com/2012/09/simple-rest-client-in-java.html
-    
+
     // To encode a query use:
     // web encoder: http://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_encodeuricomponent 
     // after that replace 3d with "="
-    
+
     public List<TransmitterRawData> readFromMongo(Context ctx, String collectionName) {
         Log.e(TAG, "starting readFromMongo");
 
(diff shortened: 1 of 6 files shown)


StephenBlackWasAlreadyTaken:
This sounds like a good solution to me.
I would much prefer Mongo rest to just cramming it in to the db manually
(and if it didn't change protocols for Mongo 3 than why not! It's a fairly
easy setup)

Will give this a go later today

tzachi-dar:
This also seems to me like a good step.
We will need to also have UI for this.
I'll also add a get function that will replace the get that the wifi thread was doing.

tzachi-dar:
We should probably also make our UI for direct mongo uploading gray, since it will stop working soon.

StephenBlackWasAlreadyTaken:
Someone may be rolling their own Mingo rather than using Mongo labs

tzachi-dar:
Didn't really think about creating my own mongodb. This could also have been a solution...

AdrianLxM:
> Someone may be rolling their own Mingo rather than using Mongo labs

For Mongo 3.0 we unfortunately won't have any solution - until a driver comes out. (Please correct me if I'm wrong)
The only thing I can think of, is make the baseURL also configurable. Then people can run their own REST server... overkill?

The ones running their own Mongo-Server also don't have to update to mongo 3.0. Version 2.6 still gets maintenance updates.
I like @tzachi-dar 's idea to grey out the "Mongo Legacy upload" but still have it as an option. People running their own mongo-server usully don't have problems configuring things in a deeper android setting.

StephenBlackWasAlreadyTaken:
Yeah, I 100% agree with that, calling it Mongo Legacy Upload makes sense, I also think we should call it Mongo API instead of Mongo REST (just to be less confusing while people are being forced to use rest)

tzachi-dar:
I don't think that changing the base url is enough for people to use other servers. for example azure have their own secret format.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoLabRest.java
```diff
@@ -0,0 +1,234 @@
...
+ * Created by adrian on 24/09/15.*
+ */
+public class MongoLabRest {
+
+    //*********************************************************************
+    // FOR TESTING:
+    // Should be read from settings/preferences in final version
+    private static final String DATABASE = "mydb";
+    private static final String API_KEY = ;
+    // How to get an API-Key: http://docs.mongolab.com/data-api/#authentication
```
StephenBlackWasAlreadyTaken, line 52: This for me was the hardest part of setup :-D

AdrianLxM:
:+1: on "Mongo API"

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -86,8 +86,19 @@ public boolean upload(List<BgReading> glucoseDataSets, List<Calibration> meterRe
 
             if (enableMongoUpload) {
```
StephenBlackWasAlreadyTaken, line 87: Im an idiot and didnt realize why it wasnt uploading... (I thought you had hardcoded it to just upload no matter what woooops!)

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoLabRest.java
```diff
@@ -0,0 +1,378 @@
...
+/**
+ * Created by adrian on 24/09/15.*
+ */
+public class MongoLabRest {
+
+    //*********************************************************************
+    // FOR TESTING:
+    // Should be read from settings/preferences in final version
+    private static final String DATABASE = "mydb";
+    private static final String API_KEY = "D2a6iaurh-oihXrraOquZSySx9QnT_Gs";
```
StephenBlackWasAlreadyTaken: Im guessing you didnt mean to check this in

app/src/main/res/xml/pref_data_sync.xml
```diff
@@ -20,14 +20,34 @@
...
                 android:inputType="textUri">
             </EditTextPreference>
+            
+            <EditTextPreference
+                android:dependency="cloud_storage_mongodb_enable"
+                android:title="key for mongolab rest api"
+                android:key="cloud_storage_mongodb_rest_key"
+                android:dialogTitle="@string/pref_dialog_mongodb_collection"
+                android:dialogMessage="This is the key that identifies your rest api"
+                android:defaultValue="D2a6iaurh-oihXrraOquZSySx9QnT_G">
```
StephenBlackWasAlreadyTaken, line 39: same

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoLabRest.java
tzachi-dar: sure, I'll remove it soon. this is  not used any more.

app/src/main/res/xml/pref_data_sync.xml
tzachi-dar, line 39: Here, I did mean to check it in. This is not the same string, and I wanted to give it as an example of what people will want to give.

AdrianLxM:
I'll close this pr and delete (=hide, restorable) the branch as the upload now works with the driver again.

tzachi-dar:
How can i find this branch?
בתאריך 18 באוק 2015 17:31,‏ "AdrianLxM" notifications@github.com כתב:

AdrianLxM:
@tzachi-dar I just restored the branch

PR #158, opened 2015-09-29
Sloppy but compiled to support mongo 3

StephenBlackWasAlreadyTaken:


tzachi-dar:
Does this work as is on android and mongo 3.0?

StephenBlackWasAlreadyTaken:
Sure does!

StephenBlackWasAlreadyTaken:
can someone test this on mongo2, I am not running into any issues on mongo3 currently

tzachi-dar:
already did a sanity check, it did work.

StephenBlackWasAlreadyTaken:
great! thanks @tzachi-dar , let me know if you think we are good to merge this for now, obviously we want to do some more testing over time but I would like to get a new build up before the forced cutover

tzachi-dar:
I'm writing a new record every 3 seconds. Already have 1200 new ones. I
would like to have a few 10,000but that is not possible at the moment. In
order to simulate a week, we need about 2100 so this is one more hour. Can
you wait that long?

StephenBlackWasAlreadyTaken:
Certainly!
Thanks for testing!!

tzachi-dar:
Ok, so far things are going great. I think you can merge.

It takes 7-20 seconds for each record, but this is not a problem for us,
and I also guess that this is very likely busy servers/network than bad
drivers.

Thanks
Tzachi

StephenBlackWasAlreadyTaken:
Thanks @tzachi-dar !!

tzachi-dar:
Now we need a new build...
בתאריך 30 בספט 2015 01:04,‏ "Stephen Black" notifications@github.com כתב:

AdrianLxM:
Testing the gap sync: After more than an hour (so more than 720 real records + 21 calibrations + 21 meter readings) I'm still below 400kB.

PR #159, opened 2015-09-30
Force the use of the dot when speaking the bg reading in mmol/l format.

JacoCronje:
On my Nexus 7 TTS works well, it shows and speaks a dot (5.4) instead of a comma (5,4). On my S5 mini it shows a comma (5,4) and the TTS fails to speak the comma. It says "five four" instead of "five point four". I guess it's got something to do with the locale settings on the devices. Anyway, I don't think the comma will ever be spoken by the TTS.

In this PR I simply forced the locale to be US such that the text to be spoken is always using a dot in order to make sure I hear "five point four" instead of "five four". 

It might also be worthwhile changing the display of "5,4" to "5.4". Maybe other people like seeing a comma instead of a dot, I prefer a dot.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/BgToSpeech.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/BgToSpeech.java
@@ -10,6 +10,7 @@ import android.util.Log;
 import com.eveningoutpost.dexdrip.UtilityModels.Constants;
 
 import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
 import java.util.Locale;
 
 /**
@@ -118,6 +119,7 @@ public class BgToSpeech {
                 df.setMaximumFractionDigits(0);
                 text =  df.format(value);
             } else {
+                df = new DecimalFormat("##.#", new DecimalFormatSymbols(Locale.US));
                 df.setMaximumFractionDigits(1);
                 df.setMinimumFractionDigits(1);
                 text =  df.format(value* Constants.MGDL_TO_MMOLL);


JacoCronje:
This change is only for the speech part, not display part. Does the French, German, ... TTS actually speak "AAA comma BBB" ? Guess I can test this by selecting other languages for TTS.

My TTS is set to English (United States), I can't find Afrikaans as an option.

AdrianLxM:
Yes, at least German will read 4.3 as "vier komma drei".

Actually, it selects the current locale of the phone as main language - and if it is not present, English as fallback. To test other languages you'd have to change the locale of the phone. (But I can confirm anyways that it reads commas in other languages).

AdrianLxM:
@JacoCronje, I changed your modification to just force the dot in the fallback case. Could you please test if that is enough?

JacoCronje:
Tested the fallback case, not working for me. It doesn't speak a comma or a dot.

JacoCronje:
The result of the setLanguage call is 0 (LANG_AVAILABLE).

09-30 15:14:55.394  12486-12486/com.eveningoutpost.dexdrip D/BgToSpeech﹕ Calling onInit()
09-30 15:14:55.394  12486-12486/com.eveningoutpost.dexdrip D/BgToSpeech﹕ status == TextToSpeech.SUCCESS
09-30 15:14:55.409  12486-12486/com.eveningoutpost.dexdrip E/BgToSpeech﹕ setLanguage result = 0

AdrianLxM:
Ok, it really looks like, your locale will write a comma as decimal separator even though it is one of the English ones - but the TTS will want a point.

Does a simple replacement work?
https://github.com/AdrianLxM/xDrip-Experimental/tree/force_dot

AdrianLxM:
The good remark from @mgranberry how to simplify BgToSpeech aside, can we close this PR as another version is already in master?

PR #161, opened 2015-10-05
Broadcast raw to NightWatch

AdrianLxM:


app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
```diff
@@ -111,6 +113,34 @@ public static void addToQueue(BgReading bgReading, String operation_type, Contex
                 bundle.putInt(Intents.EXTRA_SENSOR_BATTERY, getBatteryLevel(context));
                 bundle.putLong(Intents.EXTRA_TIMESTAMP, bgReading.timestamp);
 
+                //raw value
+                List<Calibration> cals = Calibration.latest(1);
```
StephenBlackWasAlreadyTaken: Any reason for latest instead of last?

StephenBlackWasAlreadyTaken:
This all looks great to me, merge at will!

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
AdrianLxM: last() may be faster than latest(1) - I'll change it, thanks.

PR #167, opened 2015-10-18
Bluetooth fix for Jelly Bean

timbutler:
Fix OS build matching to allow Jelly Bean based devices (which are compatible).


tzachi-dar:
I thought we already merged it.

PR #173, opened 2015-11-09
remove default creation of low alert 55 and allow modification or del…

JohanDegraeve:
…etion existing alert of this type

This is rather annoying that this type of alert can not be deleted or modified. Creation of it also makes no sense as there's already one created at 70.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
@@ -238,17 +238,6 @@ public class EditAlertActivity extends ActivityWithMenu {
             endHour = AlertType.time2Hours(at.end_time_minutes);
             endMinute = AlertType.time2Minutes(at.end_time_minutes);
             alertReraise = at.minutes_between;
-
-            if(uuid.equals(AlertType.LOW_ALERT_55)) {
-                // This is the 55 alert, can not be edited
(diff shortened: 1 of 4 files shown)


JohanDegraeve:
just noticed this on fb https://www.facebook.com/groups/cgminthecloud/permalink/857617827692808/

AdrianLxM:
I now too see the usecase for deleting the 55mg/dl alert:

If you use Share as input and want to use the more sophisticated alarms in xDrip you will have the case of the Share receiver alarming at 55 (unchangable) and xDrip alarming at the same time.

In case of a severe hypo in a public silent place like a library/classroom, ... having to disable 2 alarms on different devices while treating the hypo might actually be quite stressful.

JohanDegraeve:
well not only in combination with Share i think.
It can sometimes take time before sensor values are up again although real blood glucose values are high enough to continue. Also the silent mode override can not be disabled. 
It could be for example that I've seen just a few minutes or longer before I was getting low and took carbs, so I don't need the alert or certainly don't want to be in a situation where it starts making noise. The only thing I could do is disable low alerts for a period, or pre-snooze that specific alert, but then I need to think about, open the app etc..

jstevensog:
Agreed.  I believe the unchangeable alert should be changeable, at least to some degree.  Either the ability to make it NOT override silent mode, or set time ranges.  I know why it is there, but I never liked itbon the Dex either.

PR #174, opened 2015-11-14
Split disable alerts all into disable low and high

JohanDegraeve:
Snooze Screen : split disable alerts into 'disable all', 'disable low' and 'disable high' - also delete active bg alerts that fall within range of disabled alerts.
- The screen now has three buttons on the bottom : one to just disable low bg alerts, one for the high bg alerts and one for all alerts
- Disable and re-enable all alerts works as before, when disabling all, the two other buttons disappear. 
- When re-enabling all, the two others buttons re-appear in the screen, and also bg low and high alerts will be re-enabled, whatever the status was before
- If high alert exists, and disable high or all is done, then the existing high alert will be deleted.
- If low alert exists, and disable low or all is done, then the existing low alert will be deleted
- in the home screen, the warning messages is adapted according to the alerts being disabled : "All alerts, high alerts or low alerts"

(ignore the commit i did related to chaning warm up period of sensor - this I do during testing so I don't need to wait 2 hours)

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -767,10 +767,6 @@ public class BgReading extends Model implements ShareUploadableBg{
             Log.i("NOTIFICATIONS", "checkForRisingAllert: Notifications are currently disabled!!");
             return;
         }
-        if(prefs.getLong("high_alerts_disabled_until", 0) > new Date().getTime()){
-            Log.i("NOTIFICATIONS", "checkForRisingAllert: High alerts are currently disabled!!");
-            return;
-        }
 
         if(IsUnclearTime(context)) {
             Log.d(TAG_ALERT, "checkForRisingAllert we are in an clear time, returning without doing anything");
@@ -803,10 +799,6 @@ public class BgReading extends Model implements ShareUploadableBg{
         }
         if(prefs.getLong("alerts_disabled_until", 0) > new Date().getTime()){
             Log.d("NOTIFICATIONS", "checkForDropAllert: Notifications are currently disabled!!");
-            return;
-        }
-        if(prefs.getLong("low_alerts_disabled_until", 0) > new Date().getTime()){
-            Log.d("NOTIFICATIONS", "checkForDropAllert: Low alerts are currently disabled!!");
             return;
         }
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -296,8 +296,8 @@ public class Home extends ActivityWithMenu {
         }
 
         final long now = System.currentTimeMillis();
-        if (Sensor.currentSensor().started_at + 60000 * 5 * 2 >= now) {
-            double waitTime = (Sensor.currentSensor().started_at + 60000 * 5 * 2 - now) / 60000.0;
+        if (Sensor.currentSensor().started_at + 60000 * 60 * 2 >= now) {
+            double waitTime = (Sensor.currentSensor().started_at + 60000 * 60 * 2 - now) / 60000.0;
             notificationText.setText("Please wait while sensor warms up! (" + String.format("%.2f", waitTime) + " minutes)");
             return;
         }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -288,8 +288,8 @@ public class Home extends ActivityWithMenu {
         }
 
         final long now = System.currentTimeMillis();
-        if (Sensor.currentSensor().started_at + 60000 * 60 * 2 >= now) {
-            double waitTime = (Sensor.currentSensor().started_at + 60000 * 60 * 2 - now) / 60000.0;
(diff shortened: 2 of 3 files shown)


tzachi-dar:
On first glance looks good. I'll give it a more thorough look in the coming days.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -767,6 +767,10 @@ public static void checkForRisingAllert(Context context) {
             Log.i("NOTIFICATIONS", "checkForRisingAllert: Notifications are currently disabled!!");
             return;
         }
+        if(prefs.getLong("high_alerts_disabled_until", 0) > new Date().getTime()){
```
tzachi-dar: I don't think this is really related to the low high alert. Also this alerts is only fired once so it does not bother too much in any case.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -801,6 +805,10 @@ public static void checkForDropAllert(Context context) {
             Log.d("NOTIFICATIONS", "checkForDropAllert: Notifications are currently disabled!!");
             return;
         }
+        if(prefs.getLong("low_alerts_disabled_until", 0) > new Date().getTime()){
+            Log.d("NOTIFICATIONS", "checkForDropAllert: Low alerts are currently disabled!!");
```
tzachi-dar: Same as above.

app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
```diff
@@ -125,6 +129,16 @@ public String getMenuName() {
 
     public void addListenerOnButton() {
         buttonSnooze = (Button)findViewById(R.id.button_snooze);
+		
+		//low alerts
```
mgranberry, line 133: The whitespace appears to be off here.

tzachi-dar:
One last request:
Can you remove alerts_disabled_until from the code?
keep the 3 buttons but once the all one is pressed, simply call the disable low and disable high.

Thanks
Tzachi

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
JohanDegraeve: ok I will remove that check, rise alert indeed is something different than high alert

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
JohanDegraeve: ok I will remove that check, drop alert indeed is something different than low alert

app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
JohanDegraeve, line 133: will remove that

JohanDegraeve:
in the current implementation alerts_disabled_until disables really all : low , high , drop , rise, calibration alerts (if i'm not mistaking)

Do you propose then that only low and high alerts can be disabled ?

tzachi-dar:
My mistake then, I was not aware that disable all alerts really disables all of them... I thought that it was disabling only the high/low.
So, ignore my comment about  alerts_disabled_until.
Tzachi

tzachi-dar:
By the way, not really related to this checkin but just FYI.
If you want to have a faked sensor started immediately you can simply start it with a time from the past, no need to change anything in code.
Also, if you want to have a system for testing that gets bg going up and down, you can simply switch to wifi-wixel mode.
Than in code, replace the function readDataFake as readData. it will than give you a system that has readings going up and down.

app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
JohanDegraeve, line 133: the white space looks good in my source file

JohanDegraeve:
thanks for this info - I pshed the changes

AdrianLxM:
While you/we are at this part of xDrip: How about an "un-snooze" function? Or maybe simpler: adding "0 minutes" to the snooze time selection spinner?

JohanDegraeve:
Could be achieved with this new change by first "disable all alerts" (or disable low or high alerts), and then "re-enable all alerts".
When disabling the alerts, the active alerts are deleted (that's also a change I did)

Maybe not very clear for the end-user/ could be useful to have a "help" button which explains things a little bit

AdrianLxM:
@JohanDegraeve, I like that behaviour. Works for me :)

tzachi-dar:
Thanks for this patch.

AdrianLxM:
Before merging this I would wait for Marshmallow. It will cause merge conflicts that are easier resolved the other way around. And making xDrip usable for Marshmallow seems to be more urgent to me.

tzachi-dar:
Looks a bit too late for that (already merged).

AdrianLxM:
may I revert that?

tzachi-dar:
when do we expect to have marshmellow merged?
Can we rebase that over the master?

AdrianLxM:
I'd say @mgranberry has the lead on that ;)

mgranberry:
I just merged master back into the marshmallow branch.  No worries.

PR #175, opened 2015-11-16
Marshmellow batterywarning

AdrianLxM:
Making it possible to disable the battery warning from the settings.

@mgranberry, I'm directing this directly to the Marshmellow branch to not cause merge conflicts.


tzachi-dar:
What is the propose of this pull requests? is it for devices that don't pass the battery?

AdrianLxM:
> What is the propose of this pull requests? is it for devices that don't pass the battery?

People that use their own batteries in the transmitter have totally different voltages. With freshly inserted batteries it might already show the first warning and stay between "first warning" and "empty" for a much longer time.
Also depending on the screen the warning behind the chart/graph makes it hard to read.

tzachi-dar:
SGTM than.

PR #176, opened 2015-11-17
wakelock + time

AdrianLxM:
Added a wakelock to onHandleIntent.
The few more instructions on `ArmTimer()` were enough for my phone to decide to sleep.

Exchanged `+` for the `*` operator.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
@@ -18,6 +18,7 @@ import android.net.Uri;
 import android.os.Build;
 import android.os.Handler;
 import android.os.Looper;
+import android.os.PowerManager;
 import android.preference.PreferenceManager;
 import android.support.v4.app.NotificationCompat;
 import android.support.v4.app.NotificationManagerCompat;
@@ -92,10 +93,14 @@ public class Notifications extends IntentService {
 
     @Override
     protected void onHandleIntent(Intent intent) {
+        PowerManager pm = (PowerManager) getApplicationContext().getSystemService(Context.POWER_SERVICE);
+        PowerManager.WakeLock wl = pm.newWakeLock(PowerManager.PARTIAL_WAKE_LOCK, "NotificationsIntent");
+        wl.acquire();
         Log.d("Notifications", "Running Notifications Intent Service");
         ReadPerfs(getApplicationContext());
         notificationSetter(getApplicationContext());
         ArmTimer();
+        wl.release();
     }
 
     public void ReadPerfs(Context context) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
```diff
@@ -294,7 +299,7 @@ private void ArmTimer() {
                 AlarmManager alarm = (AlarmManager) getSystemService(ALARM_SERVICE);
                 // sleep longer if the alert is snoozed.
                 long wakeTime = activeBgAlert.is_snoozed ? activeBgAlert.next_alert_at :
-                        calendar.getTimeInMillis() + (time + 60000);
+                        calendar.getTimeInMillis() + (time * 60000);
```
tzachi-dar, line 302: nice catch.

tzachi-dar:
Are you sure that this is what has happened?
If so, let's hope that we will have time to run the wakelock itself...

AdrianLxM:
With wakelock it does not sleep on my device.
But you are right that the system still could go to sleep before the wakelock is acquired.

AFAIR, the AlarmSystem just hold a wakelock for us until `onStartComand()` returns. `onStartCommand()` is inherited from IntentService and the javaDoc advises not to.
@tzachi-dar and @mgranberry, do you know a good solution?

If you don't: Because of lack of time I would:
- first:  just merge this to have a version that "most likely" won't sleep. In case of a new value coming in, the system will stay awake for a long time anyway as all the BgSendQueue stuff is handled after this Intent is started. Maybe this is why we didn't notice it earlier.
- second: overwrite either `onStart()` or `onStartCommand()` to acquire a wakelock before doing the supercall.

(There's another phenomenon with rerise periods longer than 5 minutes or "longer then until the next value comes in" I'd like to investigate.)

mgranberry:
Thanks.

tzachi-dar:
I'll try to read about what you wrote about wake locks and see what I will
find, although I must say I have many things on my plate...

Thanks
Tzachi

mgranberry:
Approach 2 can be tricky to get right because IntentService can queue up
several intents so managing wakelocks can be difficult.

The easier correct approach is to use a broadcast PendingIntent for the
alarm and use a (WakefulBroadcastReceiver)[
http://developer.android.com/reference/android/support/v4/content/WakefulBroadcastReceiver.html]
to acquire the wakelock and release it with another broadcast to release
it.  That's a lot of work given xDrip's design, so I'm still alright with
either of Adrian's suggestions.

AdrianLxM:
There is a possibility for the System to manage the wakelocks for you in case they queue up.
You can enable "reference counting" and add/acquire a wakelock in onStartCommand. After onHandleIntent you then can query the system for one of them and release it.

PR #177, opened 2015-11-18
Marshmallow reraise

AdrianLxM:
Reraise didn't work.
- Not trigger alarm system every time a new value comes in while still in rerise period. Done this as the very last point before actually triggering the alarm to make sure all "intended side effects" of the Alarm System still work.
- Arm timer x minutes after the alarm was last active and not x minutes from now. (x = rerise value)

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
@@ -61,7 +61,9 @@ public class AlertPlayer {
         stopAlert(ctx, true, false);
         int alertIn = newAlert.minutes_between;
         if(alertIn < 1) { alertIn = 1; }
-        ActiveBgAlert.Create(newAlert.uuid, false, new Date().getTime() + alertIn * 60000 );
+        ActiveBgAlert aba = ActiveBgAlert.Create(newAlert.uuid, false, new Date().getTime() + alertIn * 60000);
+        aba.last_alerted_at = System.currentTimeMillis();
+        aba.save();
         Vibrate(ctx, newAlert, bgValue, newAlert.override_silent_mode, 0);
     }
 
@@ -121,14 +123,19 @@ public class AlertPlayer {
             stopAlert(ctx, false, false);
 
             int timeFromStartPlaying = activeBgAlert.getUpdatePlayTime();
-            AlertType alert = AlertType.get_alert(activeBgAlert.alert_uuid);
-            if (alert == null) {
+            AlertType alertType = AlertType.get_alert(activeBgAlert.alert_uuid);
+            if (alertType == null) {
                 Log.d(TAG, "ClockTick: The alert was already deleted... will not play");
                 ActiveBgAlert.ClearData();
                 return;
             }
-            Log.d(TAG,"ClockTick: Playing the alert again");
-            Vibrate(ctx, alert, bgValue, alert.override_silent_mode, timeFromStartPlaying);
+            Log.d(TAG, "ClockTick: Playing the alert again");
+            if(System.currentTimeMillis() >=  activeBgAlert.last_alerted_at + alertType.minutes_between*60000 - 30000){
+                // just alert, if the reraise time has passed (or half a minute earlier to not miss one by a few milliseconds)
+                activeBgAlert.last_alerted_at = System.currentTimeMillis();
+                activeBgAlert.save();
+                Vibrate(ctx, alertType, bgValue, alertType.override_silent_mode, timeFromStartPlaying);
+            }
         }
 
     }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
@@ -98,7 +98,7 @@ public class AlertPlayer {
     public synchronized  void PreSnooze(Context ctx, String uuid, int repeatTime) {
         Log.i(TAG, "PreSnooze called repeatTime = "+ repeatTime);
         stopAlert(ctx, true, false);
-        ActiveBgAlert.Create(uuid, true, new Date().getTime() + repeatTime * 60000 );
+        ActiveBgAlert.Create(uuid, true, new Date().getTime() + repeatTime * 60000);
         ActiveBgAlert activeBgAlert = ActiveBgAlert.getOnly();
         if (activeBgAlert  == null) {
             Log.wtf(TAG, "Just created the alert, where did it go...");
@@ -130,7 +130,7 @@ public class AlertPlayer {
             }
             Log.d(TAG, "ClockTick: Playing the alert again");
             int time = alertType.minutes_between;
-            if (time < 1) {
+            if (time < 1 || AlertPlayer.isAscendingMode(context)) {
                 time = 1;
             }
             if(System.currentTimeMillis() >=  activeBgAlert.last_alerted_at + time*60000 - 30000){
@@ -195,9 +195,13 @@ public class AlertPlayer {
         return PendingIntent.getService(ctx, 0, intent, PendingIntent.FLAG_UPDATE_CURRENT);
     }
 
+    public static boolean isAscendingMode(Context ctx){
+        return getAlertProfile(ctx) == ALERT_PROFILE_ASCENDING;
+    }
+
     static private int getAlertProfile(Context ctx){
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(ctx);
-        String profile = prefs.getString("bg_alert_profile", "ascending");
+        String profile = prefs.getString("bg_alert_profile", "High");
         if(profile.equals("High")) {
             Log.i(TAG, "getAlertProfile returning ALERT_PROFILE_HIGH");
             return ALERT_PROFILE_HIGH;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
@@ -120,11 +120,10 @@ public class AlertPlayer {
             return;
         }
         if(activeBgAlert.ready_to_alarm()) {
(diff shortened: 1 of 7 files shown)


AdrianLxM:
Tested it with artificially low high alarm the whole day. Seems to work for me.

tzachi-dar:
I will try to review it today or tomorrow.

AdrianLxM:
Disabled rerise (set to one minute) if in ascending mode.
Defaulted some settings to make sure to alarm.

(ascending mode seems to have inherited more problems that existed with rerise prior to the marshmallow adaption. I will adress them tomorrow).

tzachi-dar:
Just did a fast pass on the code here. Seems to me that this feature makes the alert system much more complicated than it was before.
There are a few fast things that I have noticed:
1) why was the defaults of alerts changed from ascending to high?
2) why was smart snoozing and smart alerting defaults changed?
3) it seems to break the layering. one place I have noticed that alertsplayer now makes changes to aba. The code should be changed so this will not happen.

I'll try to give it a more through look on the next days.

AdrianLxM:
>  1) why was the defaults of alerts changed from ascending to high?

This is one of the most common question with the alarm system: "Why does the alarm system do not play any sound?". (Especially when testing this system for the first time, people don't wait for the 3rd alarming).

>  2) why was smart snoozing and smart alerting defaults changed?

They were mistakenly switched on by default. Two undocumented features that could lead to alarms not ringing, should have to be activated by the user deliberately but not by default.

> one place I have noticed that alertsplayer now makes changes to aba. The code should be changed so this will not happen.

The time when the alarm was played has to be stored somewhere in order to know when to play it again (= change a field in aba).  There was a field for that, that was never written to, so I used it. If you want, you can find a solution with `next_alert_at` or similar. But you would also have to update and store it when an alarm sounds.

JohanDegraeve:
1) i also think it's better to set default to high. Here we've been testing the app and xdrip together with 3 T1's, it was one of the first things we all did : set it to "high". (although we used nightwatch in the beginning but it's the same behaviour there) - but maybe from practical point of view it is better to have such change in a separate dedicated branch ?
2) i don't use the smart snoozing and smart alerting, here to it looks to me better to have it unchecked by default. A user would better start without this feature, then once he/she gets to know the app, he or she can try out this feature. (with same remark maybe create a new branch 'change-default-alarm-settings')
3) is it better to add a set method to AlertPlayer that allows to modify a value ? 
last_alert_at is actually written to, in toString(), where it's actually declared.
Probably a new attribute is better ? (which would require a database update)

AdrianLxM:
>  last_alert_at is actually written to, in toString(), where it's actually declared.

I find 5 occurrences in the code where it is written to:
ActiveBgAlert.Create(); ActiveBgAlert.getUpdatePlayTime(); ActiveBgAlert.snooze(); AlertPlayer.ClockTick() and AlertPlayer.startAlert

In toString there are local/stack variables with the same name. But they are in a different scope and do not affect the class variables.

This version worked fine for me now. I'm happy if Tzachi finds another way in the next couple of days before the devices get updated to Marshmallow - but I'm good with this version.

JohanDegraeve:
ActiveBgAlert.getUpdatePlayTime() uses last_alert_at in this branch, but not in the master ?

in toString, the global variable is also used
String last_alerted_at = "last_alerted_at: " + DateFormat.getDateTimeInstance(
                DateFormat.LONG, DateFormat.LONG).format(new Date(**this.last_alerted_at**));

AdrianLxM:
@JohanDegraeve: It is used, but not changed.

tzachi-dar:
A few comments:
As for the defaults:
1) High: On betas it used to be high. People have complained that it is too high, so we have changed that to ascending in order to give them a better experience. Assuming that people don't know how to change that I believe that Ascending is more human. Another thing to remember in the case that alert is ascending it will reach full volume very fast, so it is not a problem unless one is looking at xDrip trying to get to the alert.
A change in that will also cause a change in behavior to people who did not look at that feature. Not that bad, but not that nice either.
2) I believe that for most people smart alerting is the prefered option, people don't want to wake in the middle of the night to find out that they are high but are going down and have nothing to do about it.
Since most people will not change that I believe that this is a better default for them. And again changing the defaults will cause a change for existing users.

In any case, if we do want to change:
probably in a new pull request.
Make sure we change correctly, that is not only in the UI but also every time this variables are used.
Maybe we should ask in gitter and see what other people think.

AdrianLxM:
> . Another thing to remember in the case that alert is ascending it will reach full volume very fast, so it is not a problem unless one is looking at xDrip trying to get to the alert.

With the current bugs in master/beta, "very fast" being 5 minutes or with a package drop easily 10 minutes or more.
At least this fix (setting the rerise to 1 on ascending) should be merged into master/beta quickly.

tzachi-dar:
Is there any reason that we are not just updating next_alert_at (that is when reraise is bigger, use it for bigger time?).

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
```diff
@@ -60,8 +59,10 @@ public synchronized  void startAlert(Context ctx, boolean trendingToAlertEnd, Al
 
         stopAlert(ctx, true, false);
         int alertIn = newAlert.minutes_between;
-        if(alertIn < 1) { alertIn = 1; }
-        ActiveBgAlert.Create(newAlert.uuid, false, new Date().getTime() + alertIn * 60000 );
+        if(alertIn < 1 || AlertPlayer.isAscendingMode(ctx)) { alertIn = 1; }
+        ActiveBgAlert aba = ActiveBgAlert.Create(newAlert.uuid, false, new Date().getTime() + alertIn * 60000);
+        aba.last_alerted_at = System.currentTimeMillis();
+        aba.save();
```
tzachi-dar, line 65: Here you are updating the aba, not in it's file. In this case it is trivial that you could have created it with the correct value.

AdrianLxM, line 65: Create() is also called in other occasions like PreeSnooze e.g.
Not knowing the logic behind those, I just wanted to set the time the alarm last rang just in those cases, it actually rang. Feel free to change that if you are sure it doesn't cause side effects.

AdrianLxM:
> Is there any reason that we are not just updating next_alert_at (that is when reraise is bigger, use it for bigger time?).

`next_alert_at` seems to be intended for when the snooze period is over but not related to the reraise period:

``` java
    public static void ClearIfSnoozeFinished() {
        Log.d(TAG, "ActiveBgAlert ClearIfSnoozeFinished called");
        ActiveBgAlert aba = getOnly();
        if (aba != null) {
            if(new Date().getTime() > aba.next_alert_at) {
                Log.d(TAG, "ActiveBgAlert ClearIfSnoozeFinished deleting allert");
                aba.delete();
            }
        }
    }
```

This looked like `new Date().getTime() > aba.next_alert_at)` was used to determine if the snooze period is over. But now it seems that what the method does shifted away from the name over time and my assumption was not valid.
The field ´last_alerted_at´ was there anyways (in the object and in the db) so it was easiest to use it without having to fear any side effects. Feel free to add/modify the code or find a totally separate solution with `next_alert_at`.
The less I have to do things in the code and the faster it is working the better ;)

tzachi-dar:
I hope to finish it tonight...

AdrianLxM:
After a talk with a friend we remembered why it is harder to do with `next_alert_at` the way it is now:

The reraise should actually mimic the bahaviour we had with the earlier version of xDrip and in the official Dexcom receiver:
- The alarm should ring every time, the threshold is passed in the right direction (inside range -> out).
- If it stays outside or the range, the alarm should reraise after said period.

When it once enters the range again, we need to reset the time for "reraise" but not for "snooze".

tzachi-dar:
I don't think that we want to implement it this way.
In the past xDrip was implemented that way and that caused a lot of false
alarms especially with noisy sensors.

Please let me know if you disagree.

Thanks
Tzachi

AdrianLxM:
This is one of the reasons (actually the most reason) I use reraise on 
certain alarms.

If I do quick corrections I may be in range way before a standard snooze 
time is over. With Afrezza on the market I believe more and more people 
will do so.
I still want/need the alarm if I the bg rises above the threshold again.

We might introduce a criterion when to actually cancel the alert: 2 or 3 
consecutive values in range?
Or again a general setting (false by default): "Cancel alert if 3 
consecutive values are in range". <- I'd like that even more.

On 22/11/15 21:48, tzachi-dar wrote:

> I don't think that we want to implement it this way.
> In the past xDrip was implemented that way and that caused a lot of false
> alarms especially with noisy sensors.
> 
> Please let me know if you disagree.
> 
> Thanks
> Tzachi
> 
> On Sun, Nov 22, 2015 at 10:24 PM, AdrianLxM notifications@github.com
> wrote:
> 
> > After a talk with a friend we remembered why it is harder to do with
> > next_alert_at the way it is now:
> > 
> > The reraise should actually mimic the bahaviour we had with the earlier
> > version of xDrip and in the official Dexcom receiver:
> > - The alarm should ring every time, the threshold is passed in the
> >   right direction (inside range -> out).
> > - If it stays outside or the range, the alarm should reraise after
> >   said period.
> > 
> > When it once enters the range again, we need to reset the time for
> > "reraise" but not for "snooze".
> > 
> > —
> > Reply to this email directly or view it on GitHub
> > 
> > https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/177#issuecomment-158795230
> > .
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/177#issuecomment-158798551.

---

Can't remember your password? Do you need a strong and secure password?
Use Password manager! It stores your passwords & protects your account.
Check it out at http://mysecurelogon.com/password-manager

tzachi-dar:
This seems to me like a new feature all together. The way I understand
reraise was don't bother me every minute but rather every X minutes.

Your method can be supported but looks to me like a new feature all
together, and does not have anything to do with current bug.

Is that correct?

Thanks
Tzachi

AdrianLxM:
closing this in favour of https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/183

PR #179, opened 2015-11-21
Don't reraise snoozed activebgalert if more than 3 missed readings

JohanDegraeve:
When someone stops using xdrip for a period, while xdrip app is in an alert status (active snoozed alert), .. reception is lost (normal), but the alert keeps going off.

This doesn't seem useful : if there's no signal, then either app will generate missedreading alert, or the user simply stopped using the xdrip for a while.

Happened here with 4 people that I know of using xdrip 
- two parents who use the xdrip only during the night, during the day when kid is at school alarm keeps going off
- two adults (me also) in the middle of the night - disabled bluetooth because using Dexcom receiver during the night, but snoozed alarm went off at 4 in the morning

The change is not to replay a snoozed alert when no reading for more than 10 minutes

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
@@ -165,7 +165,7 @@ public class Notifications extends IntentService {
             if (activeBgAlert.uuid.equals(newAlert.uuid)) {
                 // This is the same alert. Might need to play again...
 
-                //if more than two readings missed, don't replay
+                //if more than three readings missed, don't replay
                 if ((new Date().getTime()) - (60000 * 17) - BgReading.lastNoSenssor().timestamp > 0){
                     Log.d(TAG, "FileBasedNotifications : active alert found but not replaying it because more than two readings missed :  " + newAlert.name);
                     return;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
@@ -164,6 +164,13 @@ public class Notifications extends IntentService {
 
             if (activeBgAlert.uuid.equals(newAlert.uuid)) {
                 // This is the same alert. Might need to play again...
+
+                //if more than two readings missed, don't replay
+                if ((new Date().getTime()) - (60000 * 11) - BgReading.lastNoSenssor().timestamp > 0){
+                    Log.d(TAG, "FileBasedNotifications : active alert found but not replaying it because more than two readings missed :  " + newAlert.name);
+                    return;
+                }
+
                 Log.d(TAG, "FileBasedNotifications we have found an active alert, checking if we need to play it " + newAlert.name);
                 boolean trendingToAlertEnd = trendingToAlertEnd(context, false, newAlert);
                 AlertPlayer.getPlayer().ClockTick(context, trendingToAlertEnd, EditAlertActivity.unitsConvert2Disp(doMgdl, bgReading.calculated_value));
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
@@ -166,7 +166,7 @@ public class Notifications extends IntentService {
                 // This is the same alert. Might need to play again...
 
                 //if more than two readings missed, don't replay
-                if ((new Date().getTime()) - (60000 * 11) - BgReading.lastNoSenssor().timestamp > 0){
+                if ((new Date().getTime()) - (60000 * 17) - BgReading.lastNoSenssor().timestamp > 0){
                     Log.d(TAG, "FileBasedNotifications : active alert found but not replaying it because more than two readings missed :  " + newAlert.name);
                     return;
                 }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
@@ -166,7 +166,7 @@ public class Notifications extends IntentService {
                 // This is the same alert. Might need to play again...
 
                 //disable alert on stale data
-                if(prefs.getBoolean("disable_alerts_stale_data", true)) {
+                if(prefs.getBoolean("disable_alerts_stale_data", false)) {
                     int minutes = Integer.parseInt(prefs.getString("disable_alerts_stale_data_minutes", "15")) + 2;
                     if ((new Date().getTime()) - (60000 * minutes) - BgReading.lastNoSenssor().timestamp > 0) {
                         Log.d(TAG, "FileBasedNotifications : active alert found but not replaying it because more than three readings missed :  " + newAlert.name);
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
@@ -167,7 +167,7 @@ public class Notifications extends IntentService {
 
                 //if more than three readings missed, don't replay
                 if ((new Date().getTime()) - (60000 * 17) - BgReading.lastNoSenssor().timestamp > 0){
-                    Log.d(TAG, "FileBasedNotifications : active alert found but not replaying it because more than two readings missed :  " + newAlert.name);
+                    Log.d(TAG, "FileBasedNotifications : active alert found but not replaying it because more than three readings missed :  " + newAlert.name);
                     return;
                 }
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
@@ -165,10 +165,13 @@ public class Notifications extends IntentService {
             if (activeBgAlert.uuid.equals(newAlert.uuid)) {
                 // This is the same alert. Might need to play again...
 
-                //if more than three readings missed, don't replay
-                if ((new Date().getTime()) - (60000 * 17) - BgReading.lastNoSenssor().timestamp > 0){
-                    Log.d(TAG, "FileBasedNotifications : active alert found but not replaying it because more than three readings missed :  " + newAlert.name);
-                    return;
+                //disable alert on stale data
+                if(prefs.getBoolean("disable_alerts_stale_data", true)) {
+                    int minutes = Integer.parseInt(prefs.getString("disable_alerts_stale_data_minutes", "15")) + 2;
+                    if ((new Date().getTime()) - (60000 * minutes) - BgReading.lastNoSenssor().timestamp > 0) {
+                        Log.d(TAG, "FileBasedNotifications : active alert found but not replaying it because more than three readings missed :  " + newAlert.name);
+                        return;
+                    }
                 }
 
                 Log.d(TAG, "FileBasedNotifications we have found an active alert, checking if we need to play it " + newAlert.name);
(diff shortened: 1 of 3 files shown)


AdrianLxM:
Let's see, if there will be some restructuring for the update for Marshmallow, but at the moment all the shielding conditions are evaluated inside `AlertPlayer.getPlayer().ClockTick()`. (<- the method called directly after your code). I'd put it in there to not have them all together.

I really like it, but I would like to have it depending on the interval the user put as "Minutes until alerting for missed readings" rather than 10 minutes.

JohanDegraeve:
"_I really like it, but I would like to have it depending on the interval the user put as "Minutes until alerting for missed readings" rather than 10 minutes._"

This I was also thinking first. But on the other hand the user should receive also missed readings alert (if enabled). 
The 10 (actually 11) is aligned with the home screen, if there's two readings missed, the bg amount is in red/strikethrough.

Maybe it would be interesting to be able to set the missedreadings alerts also time dependent, like during daytime I don't need that alert, during nighttime I do need it.

tzachi-dar:
First I agree that having the no-reading alert to be according to time is best. I just never was able to find the time to do it.
Second, I would make that 17 minutes.
15 minutes in order to allow for 3 packets, and the 2 extra minutes is since sometimes packets have a hard time arriving to xDrip.
I believe that 3 packets are better in order to make sure that we are not missing an alert due to some temporary missed reading.

JohanDegraeve:
Ok Changed from 11 to 17 minutes.

Can it take 2 minutes for a packet to arrive from Dexcom Transmitter to xDrip ? I was thinking dexcom transmits it once and if xDrip fails to receive it then it's missed.

tzachi-dar:
well, if you are using the "classic" xDrip receiver than you are correct.
But the xBridge has a mechanism that it might get the packets from dexcom but have the BT disconnected, so it keeps them for another 2 minutes.
On wifi wixel, internet is used to get the packets and since it might be down, connection might happen later. In this cases however the packet time is stored when it is captured, so the time differences come actually from differences in the watches...

JohanDegraeve:
ok i'm using tghe classic xdrip indeed. I hadn't understood yet what this xBridge was about to be honest

tzachi-dar:
well, xBridge is a better xDrip: that is better battery life, longer RF receiver, not having to compile it for the transmitter and also brings the raw+filtered data.
They are important in order for xDrip not to alert for lows that are caused by pressure on the transmitter.

JohanDegraeve:
and do I need to rewire (looks like there's an additional wire between wixel and hm module ? - also the resistors)
actually battery of th xdrip looks less important to me, i use 450 mAh, i lasts about 18 hours, so I know i need to charge it every night.

but "not having to compile it for the transmitter" and resending packes for 2 minutes looks interesting

tzachi-dar:
Well, the classic xBridge required people to rewire  it. now there is fw that allows for most advantages without rewiring. (battery saving will be smaller).
I believe that there are talks on gitter that it is not as stable as the classic one, but if you have more than receiver you can try it on one of them.

AdrianLxM:
As far as I understand it with Android M to have battery optimization at all. The system-alarm waking xDrip to check for lost connection (alarms) may only rise 15 minutes after the last system-alarm.

For my usecase I'd have to say:
Usually I don't care that much about missed readings. That's why the missed reading alert is set at a longer interval. Also the melody is not very loud compared to the other alerts - what I like.
But when I was high and go to bed, something is wrong. If after the snooze time there still might be the chance to be (really) high, I want to be woken up before lipolysis and/or dka can progress.

In order to still being able to set you high alarms at 100 or 110 during the night without being annoyed and still have the safety I'd suggest to make this period configurable in the settings (with preset to 17?).

tzachi-dar:
I guess that we will have to find a way to have a timer running faster than
every 15 minutes, otherwise we are in problem no matter what...

AdrianLxM:
> I guess that we will have to find a way to have a timer running faster than every 15 minutes, otherwise we are in problem no matter what...

You can set the alarm at any time you want. It will just wake 15 minutes from the last one actually triggering the Intent. So far I didn't see a problem as this just affects the missed reading alert.

On the general alarm system, Matthias programmed it that way, that if there is active alert, it will trigger the Intent at the specified time - but drain the battery of course. If there is no active alert, it is the new reading that will trigger the Notifications and there is no need to circumvent doze.

It is not a matter of finding a way or not - it is the question if we want to allow the missed reading alert being late up to 10 minutes for the advantage of not draining the battery and get into the play store.

tzachi-dar:
There are a few things to note:
1) as for power there is no issue, the phone will have to wake up every 5 minutes. The only question is whether you do it using a BT message or some other message, so saving power is not an option.

I also believe that the xBridge needs a timer once in every 25 seconds which is even harder.

2) for missed readings I don't see a problem in waiting 10 more minutes. For other alerts 15 minutes is a lot of time,

3) In order to work around the 15 minutes problem we can AFAIK: 1) create 3 programs, each will wake the phone once in 5 minutes. 2) send a dedicated message to wake the phone once in 5 minutes. 3) send a mail (or sms) to wake the phone every 5 minutes. 4) use whatever workaround people will find to solve this problem.

AdrianLxM:
There is no need to do such dirty hacks. Please read the documentation of the doze feature: http://developer.android.com/training/monitoring-device-state/doze-standby.html

As Doze affects the whole device, therefore all the other apps as well, not just xDrip: preventing the phone to enter doze mode will lead to much higher battery consumption. While there is an active alert, we have to live with that. Without an alarm or failover being triggered once, the 15 minutes latency period is never activated.

AdrianLxM:
There is no need to do such dirty hacks (and most of them wouldn't work due to doze anyways ;) ). Please read the documentation of the doze feature: http://developer.android.com/training/monitoring-device-state/doze-standby.html

As Doze affects the whole device, therefore all the other apps as well, not just xDrip: preventing the phone to enter doze mode will lead to much higher battery consumption. While there is an active alert, we have to live with that. Without an alarm or failover being triggered once, the 15 minutes latency period is never activated.

JohanDegraeve:
I need to say that I can't follow you guys anymore.
Should I do the "17 minutes" via the settings ? With a default and minimum value of 17 minutes ?

By the way, I installed this on my xDrip https://raw.githubusercontent.com/jstevensog/wixel-sdk/master/apps/xBridge2/xBridge2.wxl
and this works without adding wires. Indeed when I turn off bluetooth for 6 to 7 minutes, and then on, it still sends 1 missed reading.

AdrianLxM:
@JohanDegraeve, I would add it to the settings with 17 minutes as the default value (that's what wee agreed on?). Then everyone can change it as to his preferences.

If you want I can do this tomorrow and send a pull request to your branch.

JohanDegraeve:
yes go ahead, you'll probably to that faster than I would

tzachi-dar:
please make that 15 minutes on ui and add 2 more minutes in code...

AdrianLxM:
@JohanDegraeve: here is my pull request: https://github.com/JohanDegraeve/xDrip-Experimental/pull/1

JohanDegraeve:
I'll check it later today

JohanDegraeve:
Ok made a few changes : Moved the Stale data setting to "BG Alerts Settings" preference page - modified texts

AdrianLxM:
Does this really just supress snoozed alarms?
I thought it would supress all bg alarms to reraise - if snoozed or not.

JohanDegraeve:
Normally an alarm would already be raised before loss of reception so then it's a snoozed alert.

I tested in the master : change an alarm while in a missed signal period, to lower level than latest measurement, here also the alarm doesn't raise, waited 20 minutes. When connection is re-established, then as soon as measurement comes in, alarm is raised.
So it looks like the change only affects the snoozed alarms.

maybe i should do a renaming in the code also, disable_alerts_stale_data_minutes ==> disable_snoozed_alerts_stale_data_minutes

AdrianLxM:
We'd have to check this, but as I understand it, it would stop any (active*)  alert. No matter if it is snoozed or not. And as this just does have an effect if the value is out of range: It should diable any alert virtually.
- Up the rearaise value to test that -  or it will get annoying ;)

JohanDegraeve:
you're right , any active alert - I'll adapt the text

app/src/main/res/xml/pref_notifications.xml
```diff
@@ -38,6 +38,20 @@
                 android:title="Shortcut to Bg Level Alerts"
                 android:summary="Create a shortcut from main navigation to bg level screen"
                 android:defaultValue="false" />
+
+                <CheckBoxPreference
+                    android:key="disable_alerts_stale_data"
+                    android:title="Suppress Snoozed Alerts if missed readings"
+                    android:summary="Suppress snoozed alerts if missed readings for a predefined period"
+                    android:defaultValue="true" />
```
tzachi-dar: This feature should be false by default.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
```diff
@@ -164,6 +164,16 @@ public void FileBasedNotifications(Context context) {
 
             if (activeBgAlert.uuid.equals(newAlert.uuid)) {
                 // This is the same alert. Might need to play again...
+
+                //disable alert on stale data
+                if(prefs.getBoolean("disable_alerts_stale_data", true)) {
```
tzachi-dar: Default should be false, we better have a false alarm than miss an alarrn.

tzachi-dar:
In any case, this will only affect alerts that are already started.
So, for the good or for the bad, in the case that alert has not started, it will be raised.
This can happen for example, if you have a rule to alert if higher than 150 after 3PM. If you loose connection on 14:30, on 15:00 the alert will be raised.

So, I think this should fit at the beginning of the function after the line 

```
  BgReading bgReading = BgReading.last();
    if(bgReading == null) {
        // Sensor is stopped, or there is not enough data
        AlertPlayer.getPlayer().stopAlert(context, true, false);
        return;
    }
```

here check that bg is from a recent time.

AdrianLxM:
I totally agree that in default I would rather have one more alert than one missed. (Same goes for the no alert on noise or when trending towards range, ...). I just put it like this to have the same behaviour.

I don't understand your objective: Do you want no new alert on stale data, @tzachi-dar? 
I want to not start a new alert on stale data but have the missed readings alert, if this feature is enabled.

Also: "Disable all alerts" does not seem to disable missed reading alerts. It may be a different feature but is related in use case: When switching the phone or not using xDrip at all for some time.

tzachi-dar:
Point #1 - Well, you have a point.

Point #3 -  "Disable all alerts" should also disable misser reading alerts. Since it does not, it is a bug, need to be fixed on another pull request.

Second issue:
I indeed don't want to have an alert on stale data, and I believe that moving the check to the start of the function will achieve that goal. Do we agree?

JohanDegraeve:
Missed reading alert feature seems like a kind of background process running forever, not checking this "disable all alerts" thing, it's a small fix i think - probably if  all alerts would be disabled (in the preferences) and then the app would start clean (ie stop the app and restart) - those alarms would be suppressed. But once the process runs, that flag is ignored

what do you mean with "stale data" exactly ? I understood it's the same as missed readings but I can't follow your discussions so i think i misunderstand

tzachi-dar:
Missed reading alert should be a very small fix, just check the condition before alerting.

As for stale data: well as I understand it, we can play a BG alert for 3 reasons:
1) We have a rule, bg higher than the rule. Example: alert when higher than 150, a BG measurement has reached with value 160.

2) BG remains the same (data keeps arriving), now there is a rule with time limit. the time limit is crossed there is an alert.

3) An alert was snoozed, the time of snoozing is over, alert is raised.

The two last alerts happen since the time is passing, and we don't want to alert in both cases.
This is why I have asked to move the code change to the beginning of the function where it would stop both cases.

Hope this is clear, but don't hesitate to ask again.

tzachi-dar:
Hi,
There is still one change that I want to make, will post it tommorow.
Good for checking user inputs. By the way, we have a dialog for that allows only positive values, but what you did is also good.
I'll send my change tomorrow, and hopefully we can finish with that.

Thanks
Tzachi

PR #183, opened 2015-11-24
Marshmallow rerase2

tzachi-dar:
Here is the way, how I think that the issue should be handled, still needs much more testing.

Any feedback is welcomed.

Thanks
Tzachi

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
@@ -413,7 +413,7 @@ public class EditAlertActivity extends ActivityWithMenu {
                     Toast.makeText(getApplicationContext(), "Reraise Value must be 1 minute or greater", Toast.LENGTH_LONG).show();
                     return;
                 } else if (alertReraise >= defaultSnooze) {
-                    Toast.makeText(getApplicationContext(), "Reraise Value must less than snooze length", Toast.LENGTH_LONG).show();
+                    Toast.makeText(getApplicationContext(), "Reraise Value must be less than snooze length", Toast.LENGTH_LONG).show();
                     return;
                 }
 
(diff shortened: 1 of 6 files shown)


AdrianLxM:
First look at the code: looking good.
I will post the results of my practical tests later.

AdrianLxM:
I did find a bug:

Alarm with reraise of two minutes should trigger at minutes: 0, 2, 4, 6, 8, ....
It actually does trigger at minutes: 0, 2, 4, 7, 9, ...

The reason behind this:
Everytime `ArmTimer()` in "Notifications.java" is called, it pushes the time it will trigger again to 2 minutes from now. This will also happen at minute 5 when the next reading comes in.

Possible solution:
If `activeBgAlert.next_alert_at` now always contains the next time an alarm could needed to be triggered (except from the incoming readings that will always trigger the process), no matter if snoozed or not, couldn't we just reduce the line
`long wakeTime = activeBgAlert.is_snoozed ? activeBgAlert.next_alert_at : alert.getNextAlertTime(ctx);` to
`long wakeTime = activeBgAlert.next_alert_at;`
(or even use the field `activeBgAlert.next_alert_at` directly now).

cheers, Adrian

AdrianLxM:
And another thing: The Snooze-activity now has some "funny" behaviour:
1. Even if the alarm is snozed it shows it is "snoozed until". It is good that it shows that there is an active alert going on, but the the text should vary depending on the `is_snoozed` flag.
   "... Alert snoozed until..." -> "...Alert will be rise again at..."
2.  Disabling and Re-enabling the alert with a rerise-value of 1 or 2 doesn't let us wait until the next reading comes in for the alarm to ring again, but it will trigger earlier if the reraise period is over earlier. IMHO no problem as the last reading was bad anyways - so if you enable the alarms you can expect to hear an alarm within the next 5 minutes. -> No problems with that in "real life", just makes it harder to test. -> No need to change.

tzachi-dar:
Message about bug in 2,4,7,9... seems correct. Sending a fix soon.

AdrianLxM:
For testing I changed it and my proposed solution seems to work fine so far.

PR #186, opened 2015-11-27
Fix suppress missed reading alerts if all alarms disabled

JohanDegraeve:
with this change, when all alerts are disabled, then there will be  no missed reading alerts

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
@@ -22,6 +22,7 @@ import com.eveningoutpost.dexdrip.UtilityModels.ForegroundServiceStarter;
 import com.eveningoutpost.dexdrip.UtilityModels.Notifications;
 
 import java.util.Calendar;
+import java.util.Date;
 
 public class MissedReadingService extends IntentService {
     SharedPreferences prefs;
@@ -42,7 +43,9 @@ public class MissedReadingService extends IntentService {
         bg_missed_minutes =  Integer.parseInt(prefs.getString("bg_missed_minutes", "30"));
         otherAlertSnooze =  Integer.parseInt(prefs.getString("other_alerts_snooze", "20"));
 
-        if (bg_missed_alerts && BgReading.getTimeSinceLastReading() > (bg_missed_minutes * 1000 * 60)) {
+        if (bg_missed_alerts && BgReading.getTimeSinceLastReading() > (bg_missed_minutes * 1000 * 60)
+                &&
+            prefs.getLong("alerts_disabled_until", 0) < new Date().getTime()) {
             Notifications.bgMissedAlert(mContext);
             checkBackAfterSnoozeTime();
         } else {
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
@@ -49,7 +49,7 @@ public class MissedReadingService extends IntentService {
             Notifications.bgMissedAlert(mContext);
             checkBackAfterSnoozeTime();
         } else  {
-            long alarmIn = (prefs.getLong("alerts_disabled_until", 0) - new Date().getTime()) * 1000 * 60;
+            long alarmIn = (prefs.getLong("alerts_disabled_until", 0) - new Date().getTime());
             if (alarmIn <= 0) {
                 alarmIn = Long.parseLong(prefs.getString("bg_missed_minutes", "30"));
             }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
@@ -48,8 +48,12 @@ public class MissedReadingService extends IntentService {
             prefs.getLong("alerts_disabled_until", 0) < new Date().getTime()) {
             Notifications.bgMissedAlert(mContext);
             checkBackAfterSnoozeTime();
-        } else {
-            checkBackAfterMissedTime();
+        } else  {
+            long alarmIn = (prefs.getLong("alerts_disabled_until", 0) - new Date().getTime()) * 1000 * 60;
+            if (alarmIn <= 0) {
+                alarmIn = Long.parseLong(prefs.getString("bg_missed_minutes", "30"));
+            }
+            checkBackAfterMissedTime(alarmIn);
         }
     }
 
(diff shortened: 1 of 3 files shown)


AdrianLxM:
Looks good.

AdrianLxM:
I guess, we have to adjust the time when to check back also.
In case someone disables the alarms and when that time is over, there were no readings.

Possible solution:
- In case we do not alarm due to disabled alarms: Set timer to `prefs.getLong("alerts_disabled_until", 0)`
- In case the user reenables the alarms by button: Call this service to reset the timer after adjusting the setting `alerts_disabled_until`.

What do you think?

JohanDegraeve:
I will change it not exactly as you propose but to make sure the alert always goes off as expected

JohanDegraeve:
@AdrianLxM : change done as you proposed. Not fully tested yet.

app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
```diff
@@ -42,11 +43,17 @@ protected void onHandleIntent(Intent intent) {
...
-        if (bg_missed_alerts && BgReading.getTimeSinceLastReading() > (bg_missed_minutes * 1000 * 60)) {
+        if (bg_missed_alerts && BgReading.getTimeSinceLastReading() > (bg_missed_minutes * 1000 * 60)
+                &&
+            prefs.getLong("alerts_disabled_until", 0) < new Date().getTime()) {
             Notifications.bgMissedAlert(mContext);
             checkBackAfterSnoozeTime();
-        } else {
-            checkBackAfterMissedTime();
+        } else  {
+            long alarmIn = (prefs.getLong("alerts_disabled_until", 0) - new Date().getTime()) * 1000 * 60;
```
AdrianLxM: I tested it and it didn't work. I believe the `* 1000 * 600` is not needed here and will bring `alarmIn` too much into the future.

AdrianLxM:
I tested it again with 
`long alarmIn = (prefs.getLong("alerts_disabled_until", 0) - new Date().getTime());`
and it works fine :+1:

JohanDegraeve:
tested and looks ok, can be merged

app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
```diff
@@ -42,11 +43,17 @@ protected void onHandleIntent(Intent intent) {
...
+                &&
+            prefs.getLong("alerts_disabled_until", 0) < new Date().getTime()) {
             Notifications.bgMissedAlert(mContext);
             checkBackAfterSnoozeTime();
-        } else {
-            checkBackAfterMissedTime();
+        } else  {
+            long alarmIn = (prefs.getLong("alerts_disabled_until", 0) - new Date().getTime());
+            if (alarmIn <= 0) {
+                alarmIn = Long.parseLong(prefs.getString("bg_missed_minutes", "30"));
```
AdrianLxM: After that test, my battery ran empty very quickly.
The `* 1000 * 60` is needed in this line?!

JohanDegraeve: The behaviour should be the same as before that change (except in case all alerts are disabled off course) - correct me if I'm wrong.
Name of checkbackAfterMissedTime should actually be changed, better be checkBackAfter, because now it's only after missed time that a new check is done, it's also after alerts are re-enabled.

Is checkback after missedTime necessary ? probably not because the MissedReadingService is getting launched each time all alerts are checked (high alerts, low alerts), except again if all alerts are disabled.
You can see this in AlertType.java line 109, a call to checkifmissedreadingalert is made. That launches again the missedreadingService, and will reset the alarms in it. That's something I didn't change, but it's not very logic, although it works.

AdrianLxM:
Before it triggered every 30 (or what the user specified) MINUTES.
Not it triggeres every 30 MILLISECONDS.

I gutess it's a simple c&p thing. Just created a pr:
https://github.com/JohanDegraeve/xDrip-Experimental/pull/2

AdrianLxM:
> Is checkback after missedTime necessary ? probably not because the MissedReadingService is getting launched each time all alerts are checked (high alerts, low alerts), except again if all alerts are disabled.

The standard alarms are basically just checked when a new reading comes in (Or a snooze-time or re-raise period is over on an active alarm).
The check when a new reading comes in is less of a check (no need to check for missed readings if there is a new reading) but more important it postpones the timer in MissedReadings. When no reading comes in, it is the timer who does the check.

AdrianLxM:
Now it really works good for me. I think we can merge.

app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
```diff
@@ -42,20 +43,26 @@ protected void onHandleIntent(Intent intent) {
...
+                &&
+                prefs.getLong("alerts_disabled_until", 0) < new Date().getTime()) {
             Notifications.bgMissedAlert(mContext);
             checkBackAfterSnoozeTime();
-        } else {
-            checkBackAfterMissedTime();
+        } else  {
+            long alarmIn = prefs.getLong("alerts_disabled_until", 0) - new Date().getTime();
+            if (alarmIn <= 0) {
+                alarmIn = Long.parseLong(prefs.getString("bg_missed_minutes", "30"))* 1000 * 60;
```
tzachi-dar, line 54: I believe that here we have to adjust also for the time from the last reading. so something like:
if (alarmIn <= 0) {
-                alarmIn = Long.parseLong(prefs.getString("bg_missed_minutes", "30"))\* 1000 \* 60 -
                                BgReading.getTimeSinceLastReading()

app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
```diff
@@ -42,20 +43,26 @@ protected void onHandleIntent(Intent intent) {
         bg_missed_minutes =  Integer.parseInt(prefs.getString("bg_missed_minutes", "30"));
         otherAlertSnooze =  Integer.parseInt(prefs.getString("other_alerts_snooze", "20"));
 
-        if (bg_missed_alerts && BgReading.getTimeSinceLastReading() > (bg_missed_minutes * 1000 * 60)) {
+        if (bg_missed_alerts && BgReading.getTimeSinceLastReading() > (bg_missed_minutes * 1000 * 60)
+                &&
+                prefs.getLong("alerts_disabled_until", 0) < new Date().getTime()) {
```
tzachi-dar, line 48: I would replace the > in >= 
and the < in <=.
The reason for that is assume that we ask for a re-check in 30 minutes, and indeed we check after 30 minutes exactly, the alarm will not fire...

tzachi-dar:
2 small comments, otherwise looks like you have got all corner case covered. (I hope).

Great work guys.

app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
AdrianLxM, line 54: @tzachi-dar, you might be onto something. In this case, we had the problem already before. It is the same behaviour as before.
We could fix that in a separate pr then.

PR #188, opened 2015-11-29
Chart tooltips

AdrianLxM:
If you touch a value point in the main chart, it will show a small toast with the time and value.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -210,6 +210,7 @@ public class Home extends ActivityWithMenu {
         previewChart.setZoomType(ZoomType.HORIZONTAL);
 
         chart.setLineChartData(bgGraphBuilder.lineData());
+        chart.setOnValueTouchListener(bgGraphBuilder.getOnValueSelectTooltipListener(this));
         previewChart.setLineChartData(bgGraphBuilder.previewLineData());
         updateStuff = true;
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -210,7 +210,7 @@ public class Home extends ActivityWithMenu {
         previewChart.setZoomType(ZoomType.HORIZONTAL);
 
         chart.setLineChartData(bgGraphBuilder.lineData());
-        chart.setOnValueTouchListener(bgGraphBuilder.getOnValueSelectTooltipListener(this));
+        chart.setOnValueTouchListener(bgGraphBuilder.getOnValueSelectTooltipListener());
         previewChart.setLineChartData(bgGraphBuilder.previewLineData());
         updateStuff = true;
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -210,6 +210,7 @@ public void setColorFilter(ColorFilter cf) {}
         previewChart.setZoomType(ZoomType.HORIZONTAL);
 
         chart.setLineChartData(bgGraphBuilder.lineData());
+        chart.setOnValueTouchListener(bgGraphBuilder.getOnValueSelectTooltipListener(this));
```
mgranberry: Passing around activities is a good way to leak large amounts of RAM.  I would feel better about it if it instantiated a new anonymous listener and passed that instead.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
```diff
@@ -1,12 +1,15 @@
...
+import android.app.Activity;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.content.res.Configuration;
 import android.graphics.Color;
 import android.preference.PreferenceManager;
 import android.text.format.DateFormat;
+import android.widget.Toast;
 
+import com.eveningoutpost.dexdrip.Home;
```
mgranberry: This import is unused.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
```diff
@@ -408,4 +413,36 @@ public String unit() {
         }
 
     }
+
+    public OnValueSelectTooltipListener getOnValueSelectTooltipListener(Activity activity){
```
mgranberry: After a little more thought about this, can you do this instead?  I don't think a new Application context is necessary for a toast.

```
    public OnValueSelectTooltipListener getOnValueSelectTooltipListener(){
        return new OnValueSelectTooltipListener();
    }

    public class OnValueSelectTooltipListener implements LineChartOnValueSelectListener{

        private Toast tooltip;

        @Override
        public synchronized void onValueSelected(int i, int i1, PointValue pointValue) {
            final java.text.DateFormat timeFormat = DateFormat.getTimeFormat(context);
            //Won't give the exact time of the reading but the time on the grid: close enough.
            Long time = ((long)pointValue.getX())*FUZZER;
            if(tooltip!= null){
                tooltip.cancel();
            }
            tooltip = Toast.makeText(context, timeFormat.format(time)+ ": " + Math.round(pointValue.getY()*10)/ 10d , Toast.LENGTH_LONG);
            tooltip.show();
        }

        @Override
        public void onValueDeselected() {
            // do nothing
        }
    }
```

AdrianLxM:
I changed it how you proposed it @mgranberry.
I don't think it would affect the RAM at all (maybe GC load minimal) as we already hold a reference to said activity. But simpler is better - no need for redundant references.

mgranberry:
I agree.  It would only be 8 bytes.  But code has a way of getting copied
in this repo and it's best to keep it clean wherever possible.

PR #190, opened 2015-12-01
Check of Missed reading alert will happen based on the time from last packet.

tzachi-dar:
This fixes issues that we saw with last pull request.

Still not verified.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
@@ -47,13 +47,15 @@ public class MissedReadingService extends IntentService {
         bg_missed_minutes =  Integer.parseInt(prefs.getString("bg_missed_minutes", "30"));
         otherAlertSnooze =  Integer.parseInt(prefs.getString("other_alerts_snooze", "20"));
 
+        long now = new Date().getTime();
+
         if (bg_missed_alerts && 
                 BgReading.getTimeSinceLastReading() >= (bg_missed_minutes * 1000 * 60) &&
-                prefs.getLong("alerts_disabled_until", 0) <= new Date().getTime()) {
+                prefs.getLong("alerts_disabled_until", 0) <= now) {
             Notifications.bgMissedAlert(context);
             checkBackAfterSnoozeTime();
         } else  {
-            long alarmIn = prefs.getLong("alerts_disabled_until", 0) - new Date().getTime();
+            long alarmIn = prefs.getLong("alerts_disabled_until", 0) - now;
             if (alarmIn <= 0) {
                 alarmIn = Long.parseLong(prefs.getString("bg_missed_minutes", "30"))* 1000 * 60 - BgReading.getTimeSinceLastReading();
             }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
@@ -12,11 +12,12 @@ import android.content.SharedPreferences;
 import android.os.Build;
 import android.os.IBinder;
 import android.preference.PreferenceManager;
-import com.eveningoutpost.dexdrip.Models.UserError.Log;
 
+import com.eveningoutpost.dexdrip.Models.UserError.Log;
 import com.eveningoutpost.dexdrip.ImportedLibraries.dexcom.ReadDataShare;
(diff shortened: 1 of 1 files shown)


JohanDegraeve:
Looks better although I don't think replacing < (>) by <= (>=) will change a lot, these timestamps are up to milliseconds, chance of matching another value is extremely small.

I'll do some tests tomorrow.

tzachi-dar:
Thanks,
I believe that the chance of this happening (==) is not that small, since we set the timer to this time...
In any case, the bigger difference is taking into account the time from last packet...
By the way, we talk on https://gitter.im/nightscout/dexdrip so fill free to join.

AdrianLxM:
The chance for == is practically zero. Reading all the preferences and doing stuff ... I doubt this takes less than a millisecond. But safe is safe.

The thing that actually may cause a problem is, that you call `new Date().getTime()` twice and there is time passing in between.
If the service gets called (by a bg level alert snooze time being over eg.) just a few milliseconds before the time is over, it will not alarm. But by the time, the new timer is set, `prefs.getLong("alerts_disabled_until", 0) - new Date().getTime()` can actually be in the past.
(Reading from the preferences and instantiating a new Object takes some time.)

In order to be safe you need to evaluate a state that cannot change. E.g.: Write it into a variable and use the same time for evaluating the condition and setting the new alarm.

tzachi-dar:
Just to make sure that we are on the same page, the function getTime returns packets in ms resolution.
that said, the android clock works on resolution which is much bigger (from my test 15 ms). So, the chance of collusion is bigger.
Here is a simple test that I was running:

Log.e("TIME", "starting test");
        long start_period = new Date().getTime();
        for(int i = 0; i < 10; i++) {
          int count = 0;
          while(true) {
            long now = new Date().getTime();
            if(now > start_period) {
              Log.e("TIME", "finished perior now = " + now + " start_period = " + start_period + " count = " + count);
              start_period = now;
              break;
            }

```
      }
    }
```

and here are the results:
12-02 11:07:12.239 E/TIME    (13655): finished perior now = 1449047232251 start_period = 1449047232250 count = 0
12-02 11:07:12.259 E/TIME    (13655): finished perior now = 1449047232264 start_period = 1449047232251 count = 0
12-02 11:07:12.269 E/TIME    (13655): finished perior now = 1449047232280 start_period = 1449047232264 count = 0
12-02 11:07:12.299 E/TIME    (13655): finished perior now = 1449047232302 start_period = 1449047232280 count = 0
12-02 11:07:12.319 E/TIME    (13655): finished perior now = 1449047232322 start_period = 1449047232302 count = 0
12-02 11:07:12.329 E/TIME    (13655): finished perior now = 1449047232336 start_period = 1449047232322 count = 0
12-02 11:07:12.349 E/TIME    (13655): finished perior now = 1449047232354 start_period = 1449047232336 count = 0
12-02 11:07:12.359 E/TIME    (13655): finished perior now = 1449047232369 start_period = 1449047232354 count = 0
12-02 11:07:12.379 E/TIME    (13655): finished perior now = 1449047232384 start_period = 1449047232369 count = 0
12-02 11:07:12.389 E/TIME    (13655): finished perior now = 1449047232400 start_period = 1449047232384 count = 0

So ,you can see the minimum clock change is 15 ms.
That said, I'll update the function to get the to once.

AdrianLxM:
Yes, the resolution of the internal clock is much higher. You can get nanoseconds from it.
But from your test you see that even when you do very simple operations it will take some time visible in milliseconds. With more costly operations like reading the settings it will get longer.

Now when the Service is called a few ms before it should trigger, it will arm a timer for a few ms in the future and then still fire instead of arming a timer a few ms in the past and thereby disabling it. Thanks.

AdrianLxM:
To be really safe we'd have to pass `now` to `BgReading.getTimeSinceLastReading()` also. :/

tzachi-dar:
not realy.

"If the stated trigger time is in the past, the alarm will be triggered
immediately. If there is already an alarm for this Intent scheduled (with
the equality of two intents being defined by filterEquals(Intent)
http://developer.android.com/reference/android/content/Intent.html#filterEquals(android.content.Intent)),
then it will be removed and replaced by this one."

so, if we set it to the past, it will happen again soon, and we are ok
(Actually this was also so probably before my last change).

AdrianLxM:
I just can say from personal experience that scheduling it in the past made it not trigger on many occasions. It seems it was a bug then and got corrected by now.
As this gets updated with the play-service updates and not with the OS (except for some Chinese watches), chances are very unlikely that this will happen then.

PR #191, opened 2015-12-02
X bridge improvements

jstevensog:
Changes the way duplicate packet detection works.
Removes duplicate packet detection in onCharacteristicChanged, and relies solely on the time window of ignoring DATA packets if they are received within 60 seconds of the first.
Always sends ACK packet to bridge, even if it is a "duplicate" DATA packet.  Prevents bridge from staying awake for 2 minutes sending DATA packets that are ignored.
Includes the reduced retry time of 25 seconds if xBridge is the data source, retains 60 seconds if not.

### DIFF
--- a/app/app.iml
+++ b/app/app.iml
@@ -102,39 +102,39 @@
     </content>
     <orderEntry type="jdk" jdkName="Android API 23 Platform" jdkType="Android SDK" />
     <orderEntry type="sourceFolder" forTests="false" />
-    <orderEntry type="library" exported="" name="bson-3.1.0-rc0" level="project" />
-    <orderEntry type="library" exported="" name="answers-1.3.0" level="project" />
-    <orderEntry type="library" exported="" name="crashlytics-core-2.3.3" level="project" />
-    <orderEntry type="library" exported="" name="zxing-android-integration-2.0.0" level="project" />
-    <orderEntry type="library" exported="" name="core-3.1.0" level="project" />
-    <orderEntry type="library" exported="" name="retrofit-2.0.0-beta2" level="project" />
-    <orderEntry type="library" exported="" name="converter-gson-2.0.0-beta2" level="project" />
-    <orderEntry type="library" exported="" name="rxjava-1.0.0" level="project" />
-    <orderEntry type="library" exported="" name="crashlytics-2.5.0" level="project" />
-    <orderEntry type="library" exported="" name="guava-18.0" level="project" />
-    <orderEntry type="library" exported="" name="seeds-functional-16.0.1" level="project" />
-    <orderEntry type="library" exported="" name="seeds-strings-16.0.1" level="project" />
-    <orderEntry type="library" exported="" name="appcompat-v7-23.1.0" level="project" />
-    <orderEntry type="library" exported="" name="usb-serial-for-android-v010" level="project" />
-    <orderEntry type="library" exported="" name="okio-1.6.0" level="project" />
-    <orderEntry type="library" exported="" name="seeds-base-16.0.1" level="project" />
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -310,16 +315,7 @@ public void onCharacteristicChanged(BluetoothGatt gatt, BluetoothGattCharacteris
...
-                    if (!Arrays.equals(lastdata, data)) {
-                        Log.v(TAG, "broadcastUpdate: new data.");
-                        setSerialDataToTransmitterRawData(data, data.length);
-                    } else if (Arrays.equals(lastdata, data)) {
-                        Log.v(TAG, "broadcastUpdate: duplicate data, ignoring");
-                    }
-                } else if (data != null && data.length > 0) {
-                    setSerialDataToTransmitterRawData(data, data.length);
-                }
+                setSerialDataToTransmitterRawData(data, data.length);
```
tzachi-dar, line 318: you need to live here the code that checks tha data !=null and data.length >0.

tzachi-dar:
A few notes:

1) the code of wifi xBridge is not affected by this change.
2) If I understand correctly, even for xBridge, it is possible to receive the same packet twice. This is just the way that transmission works. More than that, it might also happen on two different threads.
That said there is a mechanism to avoid that also on TransmitterData create.

3) As for the times being used: If I understand correctly, retransmits can happen for 2 minutes, while here the limit is one minute, maybe that should also change to 2 minutes?

In any case, since this is a change in multi threaded code, it seems to me that it needs some more testing before putting it into master.

app/app.iml
```diff
@@ -102,39 +102,39 @@
     </content>
     <orderEntry type="jdk" jdkName="Android API 23 Platform" jdkType="Android SDK" />
```
tzachi-dar, line 103: why are all this changes needed?

It seems that no file was added / removed only order of files, but still the files don't seem to be ordered by any order?

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
jstevensog, line 318: I'll check it out.
Thanks.

---- tzachi-dar wrote ----

jstevensog:
1) great news.
2) the only reason xBridge2 will send multiple DATA packets, is that it hasn't received an ACK packet from xDrip.  It doesn't normally happen as a function of BLE.  With these changes, the worst case is it will send 1 additional packet up to 10 seconds from the first.  Previously, as xDrip would only send 1 ACK packet and ignore any "duplicate" packets, the bridge would keep sending for 2 minutes after dex packet capture.
3) As explained above, with this code it is extremely unlikely that more than 2 of the same DATA packet will be received.

Cheers

---- tzachi-dar wrote ----

app/app.iml
jstevensog, line 103: This I never understood either.  The only library I added was the pngj one.  All the others existed previously.  But every merge I tried caused a whole lot of merge conflicts with the order.  Don't know when it changed so radically, I have only tried to reduce it.

Will be pushing up a new version with the order (hopefully) fixed tonight.
Cheers

---- tzachi-dar wrote ----

jstevensog:
Closing to try a different version with less conflicts.

app/app.iml
AdrianLxM, line 103: @jstevensog, simply don't call `git add` on that file. It will be created/updated by IntelliJ automatically when importing the project.

PR #193, opened 2015-12-02
X bridge improvements

jstevensog:
Re-instated data arrray null/length checks in onCharacteristicChanged as suggested by tzachi-dar.
Resolved many of the merge conflicts  (hopefully all) in app\app.iml.
This should be a better merge.


tzachi-dar:
Thanks for your changes. 
I believe that app.iml file has not changed at all, why are you changing the order of files there? (simply remove it from the checkin).

AdrianLxM:
@tzachi-dar, this is an internal file of IntelliJ. It does that rearrangement automatically. We don't need to look at that.
I still wouldn't remove it, just not add it to the commit.

tzachi-dar:
yes, I meant remove it from the commit...

jstevensog:
When I remove it from the check-in I get complaints.  Will try again.  I never changed the order and have no idea why this became an issue.
Cheers 

---- tzachi-dar wrote ----

jstevensog:
So, would it be easier to avoid this stuff in future to simply add it to .gitignore?
Cheers

---- tzachi-dar wrote ----

> yes, I meant remove it from the commit...
>

JohanDegraeve:
I never add app.iml before committing so I just keep my own version. Sometimes I need to stash it.

jstevensog:
Hopefully this commit takes care of it.  And includes it in .gitignore.  If it is not a source file it should be ignored.
Cheers

PR #196, opened 2015-12-02
First draft on using upsert for dexcom on sgv entries.

tzachi-dar:

Please check and see if this direction is good, and if it is I'll do all other objects.
Also do the same for NS api.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -364,6 +364,7 @@ public class BgReading extends Model implements ShareUploadableBg{
             bgReading.perform_calculations();
             context.startService(new Intent(context, Notifications.class));
             BgSendQueue.addToQueue(bgReading, "create", context);
+            BgSendQueue.addToQueue(bgReading, "create", context);
         }
 
         Log.i("BG GSON: ",bgReading.toS());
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -364,6 +364,7 @@ public static BgReading create(double raw_data, double filtered_data, Context co
             bgReading.perform_calculations();
             context.startService(new Intent(context, Notifications.class));
             BgSendQueue.addToQueue(bgReading, "create", context);
+            BgSendQueue.addToQueue(bgReading, "create", context);
```
AdrianLxM, line 367: Why adding it twice?

AdrianLxM:
I think `sysTime` has to be added to the actual data as well, not just in the query.

I'm not sure if it is better to use `update()` or `updateOne()`.
In case of the upsert being an update, it would be more performant. But the general case would be an insert anyways.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
tzachi-dar, line 367: Just to make sure that I'm inserting sgvs twice. Soon will be removed...

tzachi-dar:
sysTime is indeed being set in the actual data. look again.

I believe that the function that I was using is updateOne since the last parameter there is multi and I set that to false.
There are three things that make me wander however:
1) should we add also a key to the db, to make this queries faster?
2) if yes, how do we make sure it is done once and not with every packet?
3) how do we make that using the rest api?

Thanks
Tzachi

AdrianLxM:
My bad. the blank line in the github diff threw me off in the middle of the night.
Also: The method `update` seems to be the correct one now, so your code looks good.

I wouldn't add anything else new by now. To be honest, I didn't look into how mongo internally works.
On other db's, making "systime and type" key would require them to always be present. What they aren't for other and older uploaders.
As far as I understood Jason, an extra ID field wouldn't be of any use with upserts over the NS-REST-API.

By now, If I understood Jason right, we would be good if we just added the "sysTime" field exactly alike with the REST upload. (And for all uploaded types).

PR #197, opened 2015-12-03
Just shutdown tts if it exists

AdrianLxM:
Shutdown TTS just/only if it exists.


JohanDegraeve:
this fixes the problem, app does not crash anymore

AdrianLxM:
Sorry it's late: "app does not restart anymore" means it doesn't crash in the first place, I hope.

JohanDegraeve:
correct , app does not crash anymore :+1:

AdrianLxM:
Thanks, I will merge.

PR #202, opened 2015-12-04
Home Improvement

AdrianLxM:
- setupCharts() will now be called just once.
- Make some methods private to be sure they are not called from elsewhere (sure there are no side effects)
- Moved some stuff that only needs to be initialized once after installation to xdrip.java:
  - Default values for preferences
  - Start of CollectionService (on reboot it will be done by AutoStart.java)
  - IdempotentMigrations

_More power!_


app/src/main/java/com/eveningoutpost/dexdrip/xdrip.java
```diff
@@ -15,5 +19,12 @@
     public void onCreate() {
         super.onCreate();
         Fabric.with(this, new Crashlytics());
+        CollectionServiceStarter collectionServiceStarter = new CollectionServiceStarter(getApplicationContext());
```
mgranberry, line 22: It's a bit pedantic, but 'getApplicationContext()' is a bit redundant when inside an application 'Context'.  'this' would suffice.

PR #204, opened 2015-12-05
Send queue small changes

tzachi-dar:
just simple code refactoring.


AdrianLxM:
I like the renaming as the method did so much more than adding it to the queue :)

PR #205, opened 2015-12-05
Btscan wifi bt wixel fix

AdrianLxM:
Add BTScan to NavDrawer for BtWifiWixel
Don't switch to pure BTWixel mode when in BtWifiWixel-mode and BTWixel is found.


tzachi-dar:
Looks great.

PR #206, opened 2015-12-06
Create index for bgreadings(timestamp)

AdrianLxM:
@mgranberry: This would be an index on  bgreadings(timestamp) like you suggested. Does it look ok?

I thought the installation would be a good place (at least to test it). Another place I thought of were the db migrations. But that would increase the schema version number and make db's incompatible for export/reimport.

So far it at least didn't have a negative effect that I could notice.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/xdrip.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/xdrip.java
@@ -1,8 +1,10 @@
 package com.eveningoutpost.dexdrip;
 
 import android.app.Application;
+import android.database.sqlite.SQLiteDatabase;
 import android.preference.PreferenceManager;
 
+import com.activeandroid.Cache;
 import com.crashlytics.android.Crashlytics;
 import com.eveningoutpost.dexdrip.UtilityModels.CollectionServiceStarter;
 import com.eveningoutpost.dexdrip.UtilityModels.IdempotentMigrations;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/xdrip.java
```diff
@@ -19,12 +21,16 @@
...
         PreferenceManager.setDefaultValues(this, R.xml.pref_general, false);
         PreferenceManager.setDefaultValues(this, R.xml.pref_data_sync, false);
         PreferenceManager.setDefaultValues(this, R.xml.pref_notifications, false);
         PreferenceManager.setDefaultValues(this, R.xml.pref_data_source, false);
-        new IdempotentMigrations(getApplicationContext()).performAll();
+        new IdempotentMigrations(this).performAll();
+
+        //create index
+        SQLiteDatabase db = Cache.openDatabase();
+        db.execSQL("CREATE UNIQUE INDEX IF NOT EXISTS bgtimestampindex ON bgreadings(timestamp)");
```
mgranberry, line 34: You might want to experiment with `CREATE UNIQUE INDEX IF NOT EXISTS bgtimestampindex ON bgreadings(timestamp DESC)` as well.  I'm not sure which will perform better, if either.  I guess a separate index to get it going both ways is an option too.

mgranberry:
This is exactly the kind of thing that ItempotentMigrations are intended to handle.  You might put it in there.

tzachi-dar:
A few things:

1) on my code, this key is there, or am I missing something?

```
@Expose
@Column(name = "timestamp", index = true)
public long timestamp;
```

" db migrations. But that would increase the schema version number and make
db's incompatible for export/reimport."
2) Not sure why putting it in code and not in db migrations will help.
Assuming that there is a schema change, we are going to have 2 different
schemas with another number which is a bad idea.
Also, assuming one has an old backup file, why will he be able to load it
on schema that was changed in one way and not in the other?

And a few other questions regarding this:
1) Did you see a performance increase using this version?
2) (not realy related but anyway) what about deleting old transmiterdata
and bgsendqueue do they make db queries faster?

Thanks
Tzachi

On Sun, Dec 6, 2015 at 7:41 AM, Matthias Granberry <notifications@github.com

AdrianLxM:
@tzachi-dar, this is not a pr I want to have in master. Just one to toy around with and test if there are performance changes.

Thanks for the AA configuration. I will still do the tests in code to be more certain what actually happens on the db.

All tables per se stay the same - so no need to increase the number.

PR #207, opened 2015-12-09
First version of a library that reads data from NS.

tzachi-dar:
Any comments are welcomed.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/INsRestApi.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/INsRestApi.java
@@ -1,0 +1,39 @@
+package com.eveningoutpost.dexdrip.Services;
+
+import java.util.List;
+
+import retrofit.Call;
+import retrofit.http.GET;
+import retrofit.http.Query;
+import retrofit.http.Headers;
+import retrofit.http.Header;
+
+
+import com.eveningoutpost.dexdrip.Services.NsRestApiReader.NightscoutBg;
+import com.eveningoutpost.dexdrip.Services.NsRestApiReader.NightscoutMbg;
+
+public interface INsRestApi {
+    
+    // gets all sgvs
+    @GET("/api/v1/entries.json?find[type][$eq]=sgv")
+    Call<List<NightscoutBg>> getSgv(
+            @Header("Accept") String Accept,
+            @Query("find[date][$gt]") long date,
+            @Query("count") long count
+    );
+    
+    @GET("/api/v1/entries.json?find[type][$eq]=cal")
+    Call<List<NightscoutBg>> getCal(
+            @Header("Accept") String Accept,
+            @Query("find[date][$gt]") long date,
+            @Query("count") long count
+    );
+    
+    @GET("/api/v1/entries.json?find[type][$eq]=mbg")
+    Call<List<NightscoutMbg>> getMbg(
+            //@Header("Authorization") String authorization
+            @Header("Accept") String Accept,
+            @Query("find[date][$gt]") long date,
+            @Query("count") long count
+    );
+}    
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/NsRestApiReader.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/NsRestApiReader.java
@@ -1,0 +1,137 @@
+package com.eveningoutpost.dexdrip.Services;
+
+import java.io.IOException;
+import java.util.List;
+
+import com.eveningoutpost.dexdrip.Models.UserError.Log;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+
+import retrofit.Call;
+import retrofit.Response;
+import retrofit.http.GET;
+import retrofit.http.Query;
+import retrofit.Retrofit;
+import retrofit.GsonConverterFactory;
+
+import com.squareup.okhttp.logging.HttpLoggingInterceptor;
+import com.squareup.okhttp.OkHttpClient;
+
+
+public class NsRestApiReader {
+
+	class NightscoutBg {
+	    double xDrip_raw; // raw_data
+	    double xDrip_filtered; // filtered_data;
+	    Long date; // timestamp
+	    double sgv; // calculated_bg
+	}
+
+	class NightscoutMbg {
+	    Long date; // timestamp
+	    double mbg; // calculated_bg
+	    double xDrip_slope;
+	    double xDrip_intercept;
+	    double xDrip_estimate_raw_at_time_of_calibration;
+	    double xDrip_slope_confidence;
+	    double xDrip_sensor_confidence;
+	    double xDrip_raw_timestamp;
+	}
+	
+	
+    private final static String TAG = NsRestApiReader.class.getName();
+
+	
+	private INsRestApi CreateNsMethods(String baseUrl) {
+        Retrofit retrofit;
+
+        HttpLoggingInterceptor logging = new HttpLoggingInterceptor();  
+        // set your desired log level
+        logging.setLevel(HttpLoggingInterceptor.Level.BODY);
+
+        OkHttpClient httpClient = new OkHttpClient();  
+        // add your other interceptors ...
+
+        // add logging as last interceptor
+        httpClient.interceptors().add(logging);  // <-- this is the important line for logging
+
+        Gson gson = new GsonBuilder().create();
+        retrofit = new Retrofit.Builder()
+        .baseUrl(baseUrl)
+        .addConverterFactory(GsonConverterFactory.create(gson))
+        .client(httpClient)
+        .build();
+
+        return retrofit.create(INsRestApi.class);
+    }
+    
+    
+
+
+    public List<NightscoutMbg> readCalDataFromNs(String baseUrl, String key, long startTime, long maxCount) {
+        INsRestApi methods = CreateNsMethods(baseUrl);
+        List<NightscoutMbg> nightscoutMbgs = null;
+        try {
+        
+            Call<List<NightscoutMbg>> call = methods.getMbg(key,startTime, maxCount); 
+            
+            Response<List<NightscoutMbg>> response = call.execute();
+            if(response == null) {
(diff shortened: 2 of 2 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Services/INsRestApi.java
```diff
@@ -0,0 +1,39 @@
...
+    // gets all sgvs
+    @GET("/api/v1/entries.json?find[type][$eq]=sgv")
+    Call<List<NightscoutBg>> getSgv(
+            @Header("Accept") String Accept,
+            @Query("find[date][$gt]") long date,
+            @Query("count") long count
+    );
+    
+    @GET("/api/v1/entries.json?find[type][$eq]=cal")
+    Call<List<NightscoutBg>> getCal(
```
AdrianLxM, line 26: Should a "cal" be mapped to a "NightscoutBg"?

app/src/main/java/com/eveningoutpost/dexdrip/Services/NsRestApiReader.java
```diff
@@ -0,0 +1,137 @@
...
+public class NsRestApiReader {
+
+	class NightscoutBg {
+	    double xDrip_raw; // raw_data
+	    double xDrip_filtered; // filtered_data;
+	    Long date; // timestamp
+	    double sgv; // calculated_bg
+	}
+
+	class NightscoutMbg {
```
AdrianLxM, line 31: An "mbg" usually looks like this:
{"_id":"5665d1cedbb114d587f59648","device":"xDrip-DexcomShare","type":"mbg","date":1449513137477,"dateString":"2015-12-07T19:32:17.477+0100","mbg":143}

Rename this class to NightscoutCal?

app/src/main/java/com/eveningoutpost/dexdrip/Services/NsRestApiReader.java
```diff
@@ -0,0 +1,137 @@
...
+	    Long date; // timestamp
+	    double mbg; // calculated_bg
+	    double xDrip_slope;
+	    double xDrip_intercept;
+	    double xDrip_estimate_raw_at_time_of_calibration;
+	    double xDrip_slope_confidence;
+	    double xDrip_sensor_confidence;
+	    double xDrip_raw_timestamp;
```
AdrianLxM, line 39: Assuming this is a record to store calibrations in: This is what xDrip later would want. This mapping then should be done by the receiver-Function in xDrip that gets the data from the Service handling this library.

Nightscout doesn't usually have these fields in a calibration. A calibration looks like this: {"_id":"56641b8adbb114d587f57ab6","device":"xDrip-DexcomShare","type":"cal","date":1449401035638,"dateString":"2015-12-06T12:23:55.638+0100","slope":745,"intercept":27407,"scale":1}

This looks like the combination of a meter entry, a calibration entry and some more. A combination that is not needed in "Viewer" mode in xDrip anyways (except for the calibration graph maybe).

app/src/main/java/com/eveningoutpost/dexdrip/Services/INsRestApi.java
tzachi-dar, line 26: Probably not, should be probably called getBg.

app/src/main/java/com/eveningoutpost/dexdrip/Services/NsRestApiReader.java
tzachi-dar, line 39: All this fields are needed in xDrip viewer mode for a calibration graph.

But the real issue here is that we need to pass more fields than nightscout are currently using.
nightscout have a meter record and a calibration record that basically are (from xDrip perspective) a calibration.

So, there are a few things to consider:
1) which of them we want to expand. I was using the meter, but it might be a better idea to live meter as it is and extend the calibration.

2) we need to pass the extended fields to xDrip. So, maybe we should have 3 functions that will get nightscout only data:
getNsSgv()
getNsMeter()
getNsCalibration()
-- names are not final, just for the discussion --
and than have 2 more functions that will get the xDrip data (with all the fields in them):
getXdripSgv()
getXdripCalibration()

This will have all the fields that xDrip needs.
In the future, if more fields are needed, we can simply extend the appropriate class and they will be added automatically.

What do you think?

paolorotolo:
Also it seems you need to add `compile 'com.squareup.okhttp:logging-interceptor:2.6.0'` in build.gradle to compile. HttpLoggingInterceptor is in fact required by NsRestApiReader.

tzachi-dar:
This is correct. I have done this in my private branch, will need to do it in master as well.
PaoloRotolo does the api look fine for you? after all it only allows one to get the latest readings.

paolorotolo:
Yes, looks good! Can you actually make NsRestApiReader extend AsyncTask? In this way it will not block the main thread while reading from NS.

AdrianLxM:
> Can you actually make NsRestApiReader extend AsyncTask? In this way it will not block the main thread while reading from NS.

In the end, the library will start it's own service that will wake up every 5 minutes and will retrieve the latest data. It will then call back the app registered with the library.
If it just calls a notify event or actually pushes the data is yet to be specified.

tzachi-dar:
It seems to me that the library should be reading the data from NS.
Assuming that one wants to read the data every 5 minutes is not a must. One might want to read it once a day and others might read it on other occasions.
You can see in WixelReader.timeForNextRead() that the logic for the next reading is not simple, and also differ from wifi readers to parrot and so on.

Does the approach of 5 functions that I have proposed looks fine?

AdrianLxM:
The timing, when to trigger new Data should be part of the Library.
The Libre-Reader e.g. will not deliver the data every x minutes, but just when the User establishes an NFC connection.

I can write a proposal for an interface for that in the next couple of days.

tzachi-dar:
Seems to me that if we want to have a java library that reads from
nightscout it should be passive and not triger calls.
The part that reads from hw sources or websites should be part of xdrip.
But fill free to create another proposal so that i can see what you mean.
בתאריך 9 בדצמ 2015 21:14,‏ "AdrianLxM" notifications@github.com כתב:

paolorotolo:
We can always implement both. Some passive methods and a service you can activate with an API.

app/src/main/java/com/eveningoutpost/dexdrip/Services/NsRestApiReader.java
```diff
@@ -0,0 +1,137 @@
...
+
+	
+	private INsRestApi CreateNsMethods(String baseUrl) {
+        Retrofit retrofit;
+
+        HttpLoggingInterceptor logging = new HttpLoggingInterceptor();  
+        // set your desired log level
+        logging.setLevel(HttpLoggingInterceptor.Level.BODY);
+
+        OkHttpClient httpClient = new OkHttpClient();
```
mgranberry, line 53: This needs to be fed a cache to save data on mobile networks, and OkHttpClients are fairly expensive to create and will corrupt their state if multiple different OkHttpClients are using the same cache directory at the same time.  We should probably create a singleton or store one already configured in the `xdrip` object.

For big requests this can save megabytes.

In another app I use combined gt/lte selectors and perform multiple queries in parallel using the .enqueue() method for day/hour chunks to maximize the chance that values are in the cache but still allow changes to make it through and it works very well.

mgranberry:
BTW I have a WIP Nightscout client here: https://github.com/mgranberry/YADLA/tree/master/alrightypump-cloud/src/main/kotlin/com/kludgenics/alrightypump/cloud/nightscout

Read/write of entries to/from nightscout works great.  Writing treatments works, but reading treatments needs a little work to handle the fact that they aren't very well-structured and may map onto more than one record type.

app/src/main/java/com/eveningoutpost/dexdrip/Services/NsRestApiReader.java
tzachi-dar, line 53: Please explain:
Are you talking about caching connections? or are we talking about caching entries?

Using the same connection is very interesting. As for caching entries, this does not seem important since I read them by order, and I never read more than one entry more than once.

mgranberry, line 53: Handling gaps and backfills are where that approach fails.  If you ever decide to add treatments for some reason it would also benefit from a modest cache.

Reusing an OkHttpClient will cause new requests to use a shared connection pool for any new requests.  I believe it maintains 6 connections by default.  There is a greater benefit when using HTTP/2.0 or SSL and doing multiple back-to-back requests.  This code probably isn't likely to benefit much, but the REST uploader certainly would.  Some things like redirects to HTTPS will benefit from reused clients and from a persistent cache.

tzachi-dar, line 53: What should the cache hold? Should it hold objects that are received from
the remote server?
If yes, how does it know if they were changed or not?

Thanks
Tzachi

On Wed, Jan 6, 2016 at 4:33 PM, Matthias Granberry <notifications@github.com

tzachi-dar, line 53: one more question, how will the uploader benefit from items cache? I can
see how he can benefit from caching connections, but for objects we only
upload each object once.

Thanks
Tzachi

mgranberry, line 53: I was referring to an HTTP cache. The server generates an etag header containing a checksum of the data that it calculates that it will send.  The client then sends a request to the server with an If-None-Match header containing a set of etags that it believes may match the response coming from the server.  When the server receives a request it will generate the response and compare it to the etags in the request If-None-Match header.  If any match, the server responds with a 304 not modified response instead of sending the actual data.  The client presents the cached copy as though it came over the network.

The uploader will only benefit from cached redirects and reused connections.  The response cache won't be a benefit for most upload requests.

An Android app can configure the OkHttp cache like this:
`okHttpClient.setCache(new Cache(context.getCacheDir()), 1024 * 1024 * 50))`

50m may be overkill.  2M contains about 6m of my personal records.

PR #208, opened 2015-12-09
Examples for a DataPluginInterface

AdrianLxM:
Hi,

I'd like to propose an plugin interface for new data sources for xDrip (and probably other apps).
I'm not fixated on a certain design pattern but I generally like registering and callbacks. Some kind of more general Data-Broker might also be good but without experience I would guess it would be much more work to implement it than anyone has time at the moment?
The Interfaces in this PR are for illustration.

A defined Interface like this has the advantage from the xDrip side, that once a `DataReceiver` is implemented to map the data to the xDrip datastructure, other data retrieval methods can simply be plugged in. E.g. one that retrieves the data from a Share server or even another one for NightScout that gets the data pushed by the server rather than pulls it. The only thing that might have to change is one line of setup code, passing connection strings or similar.

From the side of re-usability: Other android programmes could easily reuse the Library. I'm thinking of an AndroidWear watchface that can show the data without the need of a connected Android phone (data via wifi, gsm, iPhone,...).

To integrate Tzachi's NightScout query methods, we could write a service that actually calls the query methods. This service would also implement `DataProvider`. For xDrip we have to find a way to map NighScout and ShareServer and... data to the xDrip datastructure anyways. This will be the next big task for any form of NightScout integration. At this point in time, writing a `DataReceiver` that does the job won't be more work but would give us much less problems in the future.

There would be different kind of flexibility we would have to choose between with the interface of DataReceiver. It seems to be a tradeoff between simplicity plus covering all cases now and the easy possibility to add more data types to just one library (like IOB, COB, ...) and other treatment and AP related things.

cheers, Adrian

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/DataPlugins/DataProvider.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/DataPlugins/DataProvider.java
@@ -1,0 +1,42 @@
+package com.eveningoutpost.dexdrip.DataPlugins;
+
+/**
+ * Created by adrian on 09/12/15.
+ */
+public interface DataProvider{
+
+    //TODO: write proper JavaDoc ;)
+
+    /*
+    * Register a DataReceiver that will receive callbacks when new data is available.
+    * The registration of the first DataReceiver may start background services or tasks for data retrieval.
+    * */
+    public void registerDataReceiver(DataReceiver dataReceiver);
+
+
+    /*
+    * Like registerDataReceiver(DataReceiver dataReceiver) but only expects callbacks for a certain type.
+    * This might save load as the DataProvider only has to retrieve data for types that are registered for
+    * **/
+    public void registerDataReceiver(DataReceiver dataReceiver, String type);
+
+
+    /*
+    * Register for several types of data.
+    * */
+    public void registerDataReceiver(DataReceiver dataReceiver, String[] types);
+
+
+    /*
+      Will deregister a DataReceiver.
+    * In case the count of registered receivers reaches 0, all data retrieval services or tasks
+    * that may have been started will get stopped.
+    * */
+    public void deregisterDataReceiver(DataReceiver dataReceiver);
+
+    /*
+    * Will deregister all DataReceivers.
+    * In case a Service or Background task is started it will be stopped.
+    * */
+    public void tearDown();
+}
--- a/app/src/main/java/com/eveningoutpost/dexdrip/DataPlugins/DataReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/DataPlugins/DataReceiver.java
@@ -1,0 +1,12 @@
+package com.eveningoutpost.dexdrip.DataPlugins;
+
+import android.os.Bundle;
+
+/**
+ * Created by adrian on 09/12/15.
+ */
+public interface DataReceiver {
+
+    public void onNewDataReceived(String type, Bundle data);
+
+}
--- a/app/src/main/java/com/eveningoutpost/dexdrip/DataPlugins/DataReceiverProposal2.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/DataPlugins/DataReceiverProposal2.java
@@ -1,0 +1,14 @@
+package com.eveningoutpost.dexdrip.DataPlugins;
+
+import android.os.Bundle;
+
+/**
+ * Created by adrian on 09/12/15.
+ */
+public interface DataReceiverProposal2 {
+
+    public void onNewBgData(Bundle data);
+    public void onNewCalData(Bundle data);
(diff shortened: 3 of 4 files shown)


tzachi-dar:
That was fast !!!

Here are my thoughts:

We currently have two kind of providers, one that brings only raw/filtered data (for example BT, xBridge, and the wifi wixel).
The second type, is a type that gets data bg, and calibration. These are the share and the nightwatchpro.

So we should somehow differentiate between the two types, since they have other sources of data to supply.

I believe that we can start with the second type and this way not make a revolution that is too big in xDrip.

As for the two interfaces:  I would really like to have the 3'rd one. I believe that having the list of fields that we want and can handle should be well exposed.

That said, I think that there is one more issue to note:
I believe that in order to make the implementation of this libraries simple, we should allow registration, unregistration and callbacks only in the UI thread.
Now, the problem is that all the data providers that we have are using a non ui thread to get their data.
So, we can either have each of them solve the problem (of moving the data to the main thread) alone, or we can have one broadcast receiver that will do the job.
So, I think that one broadcast receiver for all should be done, and it should get a well defined object with all the parameters. (I'm not sure that this is possible, but if not it, we should do something close as possible to this).

Does this sound reasonable?

tzachi-dar:
by the way, practically speaking, do we know someone with a share that is ready to make it a beta site (And can also help in coding/debugging?).

AdrianLxM:
I absolutely agree that this Interface would be for new DataSources that have already calculated sgv. Not for the 3 existing "xDrip as main CGM" modes (the wixels).

I also agree that xDrip is not thread safe. And it might become more of a problem now that we have combined input methods likt BT-Wixel+Wifi-Wixel running at the same time. But that is a separate issue that should be treated on the xDrip-side, not on the Library/Interface side. We should open another issue just for that.
Making everything be handled by the UI thread may work to synchronize everything. But it is discouraged and works against the principle design of Android. The BT-Callbacks that trigger new data e.g. are asynchronous callbacks.
 It will make the UI freeze while new data comes in. This would mean no scrolling in the settings or statistics while new values come in... and would lead to a very frustrating user experience.
Also, if the display is off, UI treads may be stopped by the system.
To make xDrip thread-safe, we just have to surround every statement that writes the database and every statement-sequence that read-writes the database with

```
synchronized(commonObject) {

}
```

One _common object_ that exists over all threads it the Aplication(-Context) -  xdrip.java in our case: _"Application - Base class for those who need to maintain global application state."_
But any static singleton also should do as common Object.

Just a thought about the 3rd interface design: Such a fixed interface is mostly self-explanatory what is nice. But assume, the NS-library now wants to also send IOB. We now add a method `onNewIOB()` to the interface. This would mean all other libraries would also have to change this interface to have this method. This is not a problem as long as we maintain all libraries and update that interface. No problem, If the ShareReaderLibrary e.g. does not read IOB, it simply won't call that method.
The only thing we have to ask ourselves: Could it be possible that someone else would write such a library and would give it to us pre-compiled so we cannot update it? Probably not as the library for a libre reader would still be maintained by us and be a wrapper around a BroadcastReceiver?

I don't have access to a share account. You could register for an account with a fake US address. But I think to help design the interface and what data to expect we can look at the Share2NS bridge and NightWatch. For the actual implementation I'm sure we will find someone with Share access in the NightScout universe :)

tzachi-dar:
OK, you have raised some interesting ideas here and but we are going on vacation, so it will take some time to write give a longer answer.

In any case, I believe that in order to work well, there will also be a need for some more functions, for example: a function saying what is the last packet that was received.

We will also need a function that will allow updating a calibration.

but as I said, I'll be slow on mails these days.

AdrianLxM:
Let's talk when you come back.
Enjoy your holidays @tzachi-dar!

tzachi-dar:
As for multi-threading: 
The recommendation to not work on the UI thread comes from ui experts not from SW engineers. From my experience with multi-threaded SW, it is very hard to do correctly and will make understanding the code much harder.
Using a single lock does not really solves the problem: it means that once someone will take the lock, the UI thread can not work... it does not solve anything...
By the way on xdrip today, there are only a few places that threads are being used, and they only work once in 5 minutes, so things are not that bad).
That said, I'm fine with not solving this problem right now.
(I'm even ready to wait with this problem until diabetes is cured.)

As for the other concerns, I think that our library should only be used with open source projects so we should not have any problem with other libraries. 
So, are we ok with the (high level) design?

AdrianLxM:
The UI can work when other method-execution is in an synchronized block. The only pars that do have to wait are updates to the db from the UI.
In this case it would only be the callback from the button when entering a calibration when a new value comes in.
For alerts this would be while the system checks if an alert should be risen would make the button "Snooze" in the Snooze-Activity and "save" on the EditAlerts-Activity waiting.
The user could still edit everything press the button. The system may only take a second longer to return to the home activity afterwards.

PR #209, opened 2015-12-09
xBridge battery display on Home preference

jstevensog:
Added a preference (and slightly re-ordered the preferences around Data Source for better flow) to allow people to turn on/off the xBridge battery display in Home. Default is on.
This allows those NOT using a voltage divider circuit (always 0% - Waiting for Packet in red), or those powering the bridge from a phone (always 100%) to opt to turn this display off.

I believe this is better than simply hiding the battery display if it is 0%, as it covers both cases.
Cheers

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -380,7 +380,7 @@ public class Home extends ActivityWithMenu {
 
         if (isDexbridge && displayBattery) {
             if (bridgeBattery == 0) {
-                dexbridgeBattery.setText("xBridge Battery unknown: Waiting for packet");
+                dexbridgeBattery.setText("xBridge Battery: Unknown, Waiting for packet");
                 dexbridgeBattery.setTextColor(Color.WHITE);
             } else {
                 dexbridgeBattery.setText("xBridge Battery: " + bridgeBattery + "%");
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -380,9 +380,10 @@ public class Home extends ActivityWithMenu {
 
         if (isDexbridge && displayBattery) {
             if (bridgeBattery == 0) {
-                dexbridgeBattery.setText("Waiting for packet");
+                dexbridgeBattery.setText("sBridge Battery unknown: Waiting for packet");
+                dexbridgeBattery.setTextColor(Color.WHITE);
             } else {
-                dexbridgeBattery.setText("Bridge Battery: " + bridgeBattery + "%");
+                dexbridgeBattery.setText("xBridge Battery: " + bridgeBattery + "%");
             }
             if (bridgeBattery < 50) dexbridgeBattery.setTextColor(Color.YELLOW);
             if (bridgeBattery < 25) dexbridgeBattery.setTextColor(Color.RED);
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -385,11 +385,12 @@ public class Home extends ActivityWithMenu {
             } else {
                 dexbridgeBattery.setText("xBridge Battery: " + bridgeBattery + "%");
             }
-            if (bridgeBattery < 50) dexbridgeBattery.setTextColor(Color.YELLOW);
-            if (bridgeBattery < 25) dexbridgeBattery.setTextColor(Color.RED);
+            if (bridgeBattery < 50 && bridgeBattery >25) dexbridgeBattery.setTextColor(Color.YELLOW);
+            if (bridgeBattery <= 25) dexbridgeBattery.setTextColor(Color.RED);
             else dexbridgeBattery.setTextColor(Color.GREEN);
             dexbridgeBattery.setVisibility(View.VISIBLE);
         } else {
+            dexbridgeBattery.setText("");
             dexbridgeBattery.setVisibility(View.INVISIBLE);
         }
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -385,9 +385,9 @@ public class Home extends ActivityWithMenu {
             } else {
                 dexbridgeBattery.setText("xBridge Battery: " + bridgeBattery + "%");
             }
-            if (bridgeBattery < 50 && bridgeBattery >25) dexbridgeBattery.setTextColor(Color.YELLOW);
-            if (bridgeBattery <= 25) dexbridgeBattery.setTextColor(Color.RED);
-            else dexbridgeBattery.setTextColor(Color.GREEN);
+            dexbridgeBattery.setTextColor(Color.GREEN);
+            if (bridgeBattery < 50 && bridgeBattery >30) dexbridgeBattery.setTextColor(Color.YELLOW);
+            if (bridgeBattery <= 30) dexbridgeBattery.setTextColor(Color.RED);
             dexbridgeBattery.setVisibility(View.VISIBLE);
         } else {
             dexbridgeBattery.setText("");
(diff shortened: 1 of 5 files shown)


AdrianLxM:
From pure code review it looks good to me.  :)

JohanDegraeve:
I will test it

JohanDegraeve:
I don't know exactly how to propose a pull request to this pull request so I write my comments : 

if (bridgeBattery == 0) {
                dexbridgeBattery.setText("Waiting for packet");
==> i propose to change to
**if (bridgeBattery == 0) {
                dexbridgeBattery.setText("xBridge battery level unknown, waiting for packet");**
"waiting for packet" looks rather confusing to me for the user. A user will not know this is linked to the setting "Display Bridge Battery"

and also keep the text white if bridgebattery == 0
so after the check on bridgebattery < 25 add
**if (bridgeBattery == 0) dexbridgeBattery.setTextColor(Color.WHITE);**

by the way i think the color can never be yellow. If value is eg 30, it will first be set to yellow, then to green

jstevensog:
Great feedback Johan.
I will have a go at incorporating all of that in the branch, which should
update the PR automatically (I hope).
Lunch time isn't to far away, so I might get to play with it then.
Cheers

On Thu, Dec 10, 2015 at 10:49 AM, Johan Degraeve notifications@github.com
wrote:

> I don't know exactly how to propose a pull request to this pull request so
> I write my comments :
> 
> if (bridgeBattery == 0) {
> dexbridgeBattery.setText("Waiting for packet");
> ==> i propose to change to
> 
> _if (bridgeBattery == 0) { dexbridgeBattery.setText("xBridge battery level
> unknown, waiting for packet");_
> "waiting for packet" looks rather confusing to me for the user. A user
> will not know this is linked to the setting "Display Bridge Battery"
> 
> and also keep the text white if bridgebattery == 0
> so after the check on bridgebattery < 25 add
> _if (bridgeBattery == 0) dexbridgeBattery.setTextColor(Color.WHITE);_
> 
> by the way i think the color can never be yellow. If value is eg 30, it
> will first be set to yellow, then to green
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/209#issuecomment-163439675
> .

## 

John Stevens

jstevensog:
Thinking about it some more, the "Waiting for Packet" will only ever appear
once.  Since Stephen moved the storage of the battery into a preference, it
is only ever 0 once.  Might need to rethink the whole thing.  Like, write
some code to set it to 0 if a packet hasn't been seen for >15 minutes.
Cheers

On Thu, Dec 10, 2015 at 10:49 AM, Johan Degraeve notifications@github.com
wrote:

> I don't know exactly how to propose a pull request to this pull request so
> I write my comments :
> 
> if (bridgeBattery == 0) {
> dexbridgeBattery.setText("Waiting for packet");
> ==> i propose to change to
> 
> _if (bridgeBattery == 0) { dexbridgeBattery.setText("xBridge battery level
> unknown, waiting for packet");_
> "waiting for packet" looks rather confusing to me for the user. A user
> will not know this is linked to the setting "Display Bridge Battery"
> 
> and also keep the text white if bridgebattery == 0
> so after the check on bridgebattery < 25 add
> _if (bridgeBattery == 0) dexbridgeBattery.setTextColor(Color.WHITE);_
> 
> by the way i think the color can never be yellow. If value is eg 30, it
> will first be set to yellow, then to green
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/209#issuecomment-163439675
> .

## 

John Stevens

JohanDegraeve:
with an xDrip that does not have the resistors, but with the xbridge2.wxl software installed, and configured on the app as an xbridge, if I enable the setting now "Disaplay Bridge Battery", then the red warning stays there for ever.

Personally i would makeit even more simply : as long as now bridge battery level > 0 is received, don't show the message at all, and don't show the setting at all.
As soon as a battery level > 0 is received, then show the setting (default checked), and then also show the message on the home screen according to the value of the setting.

Like this : xdrip users (ie with resistors but with xbridge2 software) will never notice, they don't need the feature.
xbridge users (ie with resistors) they will be able to use the feature

AdrianLxM:
@JohanDegraeve, how would you reset that?
Every time a new device is selected after a BT-Scan and every time the collection method is changed to xBridge?
That would be a logic that does not depend on the current state but on transitions from the state before (that you can no longer see on the interface) and therefore harder to support on facebook than a simple setting.

jstevensog:
Johan, I'll try and fix my logic.  I did it in a rush, and it seemed fine when I tried it.  That would be simplest I think.  There are people with the xDrip bridge running xBridge2 with the xDrip voltage divider resistors, which xBridge2 automatically detects and supports.

I am not sure your logic can be easily implemented to deal with all cases.
Cheers

JohanDegraeve:
I'll wait for the next update from John

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -375,19 +375,22 @@ private void displayCurrentInfo() {
...
+                dexbridgeBattery.setTextColor(Color.WHITE);
             } else {
-                dexbridgeBattery.setText("Bridge Battery: " + bridgeBattery + "%");
+                dexbridgeBattery.setText("xBridge Battery: " + bridgeBattery + "%");
             }
-            if (bridgeBattery < 50) dexbridgeBattery.setTextColor(Color.YELLOW);
-            if (bridgeBattery < 25) dexbridgeBattery.setTextColor(Color.RED);
+            if (bridgeBattery < 50 && bridgeBattery >25) dexbridgeBattery.setTextColor(Color.YELLOW);
+            if (bridgeBattery <= 25) dexbridgeBattery.setTextColor(Color.RED);
             else dexbridgeBattery.setTextColor(Color.GREEN);
```
JohanDegraeve: note that it wil still be reset to green whenver > 25, it will never be yellow
and also when bridgeBattery = 0, it's again set to Red.

jstevensog: You are right, but I had not noticed that before.  Fixed the logic now I
believe.
I cannot see a reason why you would always have the red Waiting for Packet
message showing permanently though.  I will try and run it up on my test
phone and see if I can reproduce.
Cheers

On Fri, Dec 11, 2015 at 9:27 AM, Johan Degraeve notifications@github.com
wrote:

> In app/src/main/java/com/eveningoutpost/dexdrip/Home.java
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/209#discussion_r47296192
> :
> 
> > ```
> >          }
> > ```
> > -            if (bridgeBattery < 50) dexbridgeBattery.setTextColor(Color.YELLOW);
> > -            if (bridgeBattery < 25) dexbridgeBattery.setTextColor(Color.RED);
> > -            if (bridgeBattery < 50 && bridgeBattery >25) dexbridgeBattery.setTextColor(Color.YELLOW);
> > -            if (bridgeBattery <= 25) dexbridgeBattery.setTextColor(Color.RED);
> >            else dexbridgeBattery.setTextColor(Color.GREEN);
> 
> note that it wil still be reset to green whenver > 25, it will never be
> yellow
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/209/files#r47296192
> .

## 

John Stevens

JohanDegraeve:
note that i use an xdrip without resistors but with xbridge2 software installed. I think it's normal that I see the red message because battery level stays 0

jstevensog:
Yep, I understood that.
xDrip bridge also has voltage divider resistors, and some use them with
their own modified xdrip-wixel code.
The xBridge2 code detects if the circuit is xBridge2 or xdrip-wixel and
sets up the voltage calculation of 0-100% appropriately.  It also "learns"
the limits of the individual circuit.  This is very rudimentary and has
hard limits on both in case something goes very awry.
xBridge2 voltage divider is lower current impact (better battery life), so
if you were ever going to add this to your setup, and are sticking with the
xdrip-wixel circuit, use the voltage divider resistors for xdrip-wixel, not
xBridge2.

Cheers

On Fri, Dec 11, 2015 at 10:00 AM, Johan Degraeve notifications@github.com
wrote:

> note that i use an xdrip without resistors but with xbridge2 software
> installed. I think it's normal that I see the red message because battery
> level stays 0
> 
> —
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/209#issuecomment-163779934
> .

## 

John Stevens

PR #210, opened 2015-12-09
Calibration timestamp fixes

AdrianLxM:
I hope this fixes the issue with wrong times shown in the calibration graph with ShareReceiver generated data. (I have no Share to test that.)


stavlor:
Will do a build when I get a chance and let you know if it fixes the cal graph times

PR #212, opened 2015-12-12
Edit Alert overhaul

AdrianLxM:
This PR solves 4 issues with the EditAlertActivity reported over the last couple of month:
- Threshold validity checks were only made for newly created alerts. It was possible to add a valid alert and then change the threshold to the one of another existing alert.
- A high alert and a low alert would collide and be rejected if the low alert was higher than the high alert. Even if they were at totally different times of the day. E.g. Low-Alert of 150mg/dl when doing sports in the afternoon would not be possible with a high alert of 145mg/dl just during the night. Now it checks if they are overlapping before rejecting them
- Minutes below 10 would show with just one digit. 9:05 am would show as 9:5.
- The TimePicker did always have 0:00 or 23:59 when opening it -> now it starts with the time previously selected.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/AlertList.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/AlertList.java
@@ -203,9 +203,9 @@ public class AlertList extends ActivityWithMenu {
         return "";
     }
 
-    public String timeFormatString(int Hour, int Minute) {
+    public String timeFormatString(int hour, int minute) {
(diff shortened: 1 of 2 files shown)


stavlor:
Looks good to me

JohanDegraeve:
most of these errors I had seen and this looks all fixed now

AdrianLxM:
Forgot the case where both time frames pass midnight: https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/216

PR #213, opened 2015-12-12
Marshmallow support

supersam654:
This PR is a copy of https://github.com/StephenBlackWasAlreadyTaken/xDrip/pull/122.

Sorry for the taking so long to move this over from regular to experimental. I currently don't have a CGM with me (and won't have access to one for a couple of weeks) so can someone (maybe @scottc5) re-verify that this still works.


app/src/main/java/com/eveningoutpost/dexdrip/utils/LocationHelper.java
```diff
@@ -0,0 +1,71 @@
...
+
+import com.eveningoutpost.dexdrip.R;
+
+/**
+ * Helper for checking if location services are enabled on the device.
+ */
+public class LocationHelper {
+    // Manually defining Marshmallow to maintain backwards API compatibility.
+    // http://developer.android.com/reference/android/os/Build.VERSION_CODES.html#M
+    private static final int MARSHMALLOW = 23;
```
AdrianLxM: Why can't `Build.VERSION_CODES.M` be used?
This should be set at compile time and not at runtime.

app/src/main/java/com/eveningoutpost/dexdrip/BluetoothScan.java
```diff
@@ -83,21 +81,8 @@ protected void onCreate(Bundle savedInstanceState) {
...
-
-                ActivityCompat.requestPermissions(this,
-                        new String[]{android.Manifest.permission.ACCESS_COARSE_LOCATION},
-                        0);
-
-
-            }
-        }
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP)
-            initializeScannerCallback();
```
AdrianLxM, line 99: `initializeScannerCallback()` is no longer needed?
Especially as the changes just add something just for M but take away this for Lollipop?

supersam654:
I redefined `VERSION_CODES.M` as `MARSHMALLOW` because xDrip (not experimental) targets a lower API version that isn't aware of that constant. And sorry about getting rid of the scanner callback part. Both have been fixed now.

AdrianLxM:
Thanks a lot! Apart from looking at the code I cannot test any further in lack of a MM device, sorry.

mgranberry:
Thanks.  Sorry that I didn't see this elsewhere.  I'll test it.

mgranberry:
I was able to re-pair the device with location services.  I would like to see GPS changed to location or location services instead because this doesn't actually use GPS, but it would make it possible to identify location using BLE beacons.  It is also worth mentioning in the string that xDrip does not track location and that it can be disabled again after pairing.

supersam654:
How's this look? I kept it as "location" and not "location services" because I felt like "location" was more understandable (while I think "location services" is more correct).

device-2015-12-12-123332

tzachi-dar:
Seems that on code GPS is still found many times.
Shouldn't this all be changed to location?

supersam654:
Normally I wouldn't change variable/string names for a UI change like this, but here you go,

mgranberry:
Thanks.

tzachi-dar:
Thanks for sending this.

PR #215, opened 2015-12-12
Snooze alarm until you re-enable and improvements in snoozeactivity

JohanDegraeve:
- possibility to snooze 10 years, this corresponds to for-ever. The
  corresponding text in the snoozeValue list will be "until you
  re-enable".
- Added message in snoozeactivity indicating if all, low and/or high
  alerts are disabled and for how long
- text is changed when re-enabling alerts
- all to displaystatus in a separate thread, to make sure new texts are
  always correctly shown, override onResume, also make call to
  displaystatus

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
@@ -116,15 +116,15 @@ public class SnoozeActivity extends ActivityWithMenu {
             alertStatus.setTextSize(TypedValue.COMPLEX_UNIT_SP, 30);
             buttonSnooze.setTextSize(TypedValue.COMPLEX_UNIT_SP, 30);
         }
-
-        displayStatus();
+        displayStatus(); //still needed to set margins?
     }
 
     @Override
-    protected void onResume() {
-        super.onResume();
-        //also make sure the text in the Activity is changed
-        displayStatus();
+    public void onWindowFocusChanged(boolean hasFocus) {
+        super.onWindowFocusChanged(hasFocus);
+        if (hasFocus){
+            displayStatus();
+        }
     }
 
     @Override
@@ -134,16 +134,16 @@ public class SnoozeActivity extends ActivityWithMenu {
 
     public void addListenerOnButton() {
         buttonSnooze = (Button)findViewById(R.id.button_snooze);
-		
-		//low alerts
-		disableLowAlerts = (Button)findViewById(R.id.button_disable_low_alerts);
-		clearLowDisabled = (Button)findViewById(R.id.enable_low_alerts);
-        
-		//high alerts
-		disableHighAlerts = (Button)findViewById(R.id.button_disable_high_alerts);
-		clearHighDisabled = (Button)findViewById(R.id.enable_high_alerts);
 
-		//all alerts
+        //low alerts
+        disableLowAlerts = (Button)findViewById(R.id.button_disable_low_alerts);
+        clearLowDisabled = (Button)findViewById(R.id.enable_low_alerts);
+
+        //high alerts
+        disableHighAlerts = (Button)findViewById(R.id.button_disable_high_alerts);
+        clearHighDisabled = (Button)findViewById(R.id.enable_high_alerts);
+
+        //all alerts
         disableAlerts = (Button)findViewById(R.id.button_disable_alerts);
         clearDisabled = (Button)findViewById(R.id.enable_alerts);
         buttonSnooze.setOnClickListener(new View.OnClickListener() {
@@ -222,8 +222,8 @@ public class SnoozeActivity extends ActivityWithMenu {
                         if (aba != null) {
                             AlertType activeBgAlert = ActiveBgAlert.alertTypegetOnly();
                             if (disableType.equalsIgnoreCase("alerts_disabled_until")
-                            || (activeBgAlert.above && disableType.equalsIgnoreCase("high_alerts_disabled_until"))
-                            || (!activeBgAlert.above && disableType.equalsIgnoreCase("low_alerts_disabled_until"))
+                                    || (activeBgAlert.above && disableType.equalsIgnoreCase("high_alerts_disabled_until"))
+                                    || (!activeBgAlert.above && disableType.equalsIgnoreCase("low_alerts_disabled_until"))
                                     ) {
                                 //active bg alert exists which is a type that is being disabled so let's remove it completely from the database
                                 ActiveBgAlert.ClearData();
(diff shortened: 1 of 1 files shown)


JohanDegraeve:
by teh way, there's still one thing I couldn't solve : 
If you have the snoozeactivity open, no active alerts, no alearts disabled. Just the text "No active alert exists" and the three buttons (disable low, high, all alerts).
Keep the app open on that screen.
Then there's an alert, a notification appears on top, scrol it down, select it. The snoozeativity opens, but the buttonSnooze and snoozeValue are not popping up. Something to do with the fact that the visibility is set to View.GONE

AdrianLxM:
Not even overriding `protected void onNewIntent(Intent intent)` worked for me.
What should work however is overriding `onWindowFocusChanged(boolean hasFocus)`:

``` java
    @Override
    public void onWindowFocusChanged(boolean hasFocus) {
        super.onWindowFocusChanged(hasFocus);
        if (hasFocus){
            displayStatus();        
        }
    }
```

app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
```diff
@@ -293,24 +299,57 @@ void displayStatus() {
...
+                status += "\n\nLow alerts disabled until " + textToAdd;
+            }
+            if (prefs.getLong("high_alerts_disabled_until", 0) > now) {
+                String textToAdd = (prefs.getLong("high_alerts_disabled_until", 0) > now + (infiniteSnoozeValueInMinutes - 365 * 24 * 60) * 60 * 1000)
+                        //if high alerts would have been disabled "until you re-enable", and this test is done less than 365 * 24 * 60 minutes later, then this test will give true
+                        ? "you re-enable":DateFormat.getTimeInstance(DateFormat.MEDIUM).format(new Date(prefs.getLong("high_alerts_disabled_until", 0)));
+                status += "\n\nHigh alerts disabled until " + textToAdd;
+            }
+        }
+        alertStatus.post(new Runnable() {
```
AdrianLxM: Is there any reason to post that to the element.
We are on an UI thread anyways. This could lead to lags in updating the UI (seen with the Database-Import that also posts to the UI).

AdrianLxM:
- I tested it (with "onWindowFocusChanged" and direct setText) and it works fine.
- I'm good with 10 years being set as "a very long time to snooze" and "at least one year from now" to check for it.
- An alarm that is in its "rerise" period still shows as snoozed.

I sent you a PR with my version that I tested and works for me: https://github.com/JohanDegraeve/xDrip-Experimental/pull/3

There are still some issues:
- Should we give the user a Button "re-enable" alert? For us it seems logical to simply disable/reenable all alerts.
- After snoozing "forever", the status shows "Alert snoozed until 21:14:21 (30456 minutes left)". Couldn't we change that so something more meaningful. (The 30456 isn't right anyways). This happens with your (original) version and with my update.

app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
tzachi-dar: Agree with adrian, this can be called directly.

app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
```diff
@@ -293,24 +299,57 @@ void displayStatus() {
...
+                String textToAdd = (prefs.getLong("low_alerts_disabled_until", 0) > now + (infiniteSnoozeValueInMinutes - 365 * 24 * 60) * 60 * 1000)
+                        //if low alerts would have been disabled "until you re-enable", and this test is done less than 365 * 24 * 60 minutes later, then this test will give true
+                        ? "you re-enable":DateFormat.getTimeInstance(DateFormat.MEDIUM).format(new Date(prefs.getLong("low_alerts_disabled_until", 0)));
+                status += "\n\nLow alerts disabled until " + textToAdd;
+            }
+            if (prefs.getLong("high_alerts_disabled_until", 0) > now) {
+                String textToAdd = (prefs.getLong("high_alerts_disabled_until", 0) > now + (infiniteSnoozeValueInMinutes - 365 * 24 * 60) * 60 * 1000)
+                        //if high alerts would have been disabled "until you re-enable", and this test is done less than 365 * 24 * 60 minutes later, then this test will give true
+                        ? "you re-enable":DateFormat.getTimeInstance(DateFormat.MEDIUM).format(new Date(prefs.getLong("high_alerts_disabled_until", 0)));
```
tzachi-dar, line 359: you have here the code that does the calculation of how long the alerts are disabled, 3 times.
Please move that to a single static function and call that also from the home activity that also displays, low alerts disabled...

tzachi-dar:
Disabling for ever, can make a reasonable choice for all alerts or low/high, but I don't think that it is reasonable for a specific alert.
It seems to me that we should allow this option only for these 3 cases, and if we want to disable an alert for a long period we should add an option to enable/disable each alert alone. This way the UI will show which alerts are on and which are not.

JohanDegraeve:
I will still test and see if I can include the comments

app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
JohanDegraeve, line 359: i'm not sure if this is worth putting in a function. It's just a single line of code, and in Home.java the check is slightly different

AdrianLxM, line 359: @tzachi-dar, are you thinking of a function `private String timeDisabledUntil(SharedPreferences prefs, String lookupString, long now)`? Seems like a parameter list almost as long as the code itself ;)

AdrianLxM:
> Disabling for ever, can make a reasonable choice for all alerts or low/high, but I don't think that it is reasonable for a specific alert.
> It seems to me that we should allow this option only for these 3 cases, and if we want to disable an alert for a long period we should add an option to enable/disable each alert alone. This way the UI will show which alerts are on and which are not.

There is a difference between disabling an alert and long-snoozing it. No matter how long you snooze an alert, if any other alert with higher priority will become active, this alert will get cancelled and therefore the snooze time is automatically over.

I don't see that many usecases for me for that either. But it fits right into the existing logic so adding one more time won't hurt.

JohanDegraeve:
I'm restricting the disable for ever for the disable all only. It's true it doesn't make much sense to do this for active bg alerts.
There's two users here who use the nightwatch app to monitor their kids. They prefer to be able to disable the alerts on the xdrip app, but still in exceptional cases (like during holidays) they want to use xdrip alarms, in that case they will re-enable the all's

I'm testing

tzachi-dar:
Please note that to disable bg alerts, one can also change profile to
silence... But i'm ok with other way to silence them.
בתאריך 13 בדצמ 2015 18:02,‏ "Johan Degraeve" notifications@github.com כתב:

JohanDegraeve:
disable "until you re-enable" is now only for the disable all buttons

please check

tzachi-dar:
ok, played with it a bit, looks nice.
Thank you.

PR #218, opened 2015-12-13
Fix a bug that caused timer to be fired all the time.

tzachi-dar:
This has happened in the case that no missing reading alerts has happened and no packets were coming in.
Also use a smarter condition on when to fire the alert.

Still under test.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
@@ -48,17 +48,24 @@ public class MissedReadingService extends IntentService {
         otherAlertSnooze =  Integer.parseInt(prefs.getString("other_alerts_snooze", "20"));
 
         long now = new Date().getTime();
+        
+        if (!bg_missed_alerts) {
+        	// we should not do anything in this case. Until there is a listener on UI changes, try again in 5 minutes
+            Log.d(TAG, "Setting timer to  5 minutes from now" );
+            checkBackAfterMissedTime(now + 5 * 60 *1000);
+        	return;
+        }
 
-        if (bg_missed_alerts && 
-                BgReading.getTimeSinceLastReading() >= (bg_missed_minutes * 1000 * 60) &&
+        if (BgReading.getTimeSinceLastReading() >= (bg_missed_minutes * 1000 * 60) &&
                 prefs.getLong("alerts_disabled_until", 0) <= now) {
             Notifications.bgMissedAlert(context);
             checkBackAfterSnoozeTime();
         } else  {
-            long alarmIn = prefs.getLong("alerts_disabled_until", 0) - now;
-            if (alarmIn <= 0) {
-                alarmIn = Long.parseLong(prefs.getString("bg_missed_minutes", "30"))* 1000 * 60 - BgReading.getTimeSinceLastReading();
-            }
+            
+            long disabletime = prefs.getLong("alerts_disabled_until", 0) - now;
+            
+            long missedTime = Long.parseLong(prefs.getString("bg_missed_minutes", "30"))* 1000 * 60 - BgReading.getTimeSinceLastReading();
+            long alarmIn = Math.max(disabletime, missedTime);
             Log.d(TAG, "Setting timer to  " + alarmIn / 60000 + " minutes from now" );
             checkBackAfterMissedTime(alarmIn);
         }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
@@ -16,6 +16,7 @@ import android.preference.PreferenceManager;
 import com.eveningoutpost.dexdrip.Models.UserError.Log;
 import com.eveningoutpost.dexdrip.ImportedLibraries.dexcom.ReadDataShare;
 import com.eveningoutpost.dexdrip.Models.BgReading;
+import com.eveningoutpost.dexdrip.Models.UserNotification;
 import com.eveningoutpost.dexdrip.Sensor;
 import com.eveningoutpost.dexdrip.UtilityModels.AlertPlayer;
 import com.eveningoutpost.dexdrip.UtilityModels.CollectionServiceStarter;
@@ -27,7 +28,7 @@ import java.util.Date;
 
 public class MissedReadingService extends IntentService {
     int otherAlertSnooze;
-    private final static String TAG = AlertPlayer.class.getSimpleName();
+    private final static String TAG = MissedReadingService.class.getSimpleName();
 
     public MissedReadingService() {
         super("MissedReadingService");
@@ -57,7 +58,7 @@ public class MissedReadingService extends IntentService {
         if (BgReading.getTimeSinceLastReading() >= (bg_missed_minutes * 1000 * 60) &&
                 prefs.getLong("alerts_disabled_until", 0) <= now) {
             Notifications.bgMissedAlert(context);
-            checkBackAfterSnoozeTime();
+            checkBackAfterSnoozeTime(now);
         } else  {
             
             long disabletime = prefs.getLong("alerts_disabled_until", 0) - now;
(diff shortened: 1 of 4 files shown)


AdrianLxM:
Indeed, it looks like this can happen, if the missed reading alerts are disabled, the time will have passed: we will fire the Service over and over again in a negative future time (therefore now).

Instead of having the Service fire every 5 minutes we should either listen for changes in the preferences. It's a simple logic: "If the setting has changed, start the MissedReadingService"

app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
```diff
@@ -48,17 +48,24 @@ protected void onHandleIntent(Intent intent) {
         otherAlertSnooze =  Integer.parseInt(prefs.getString("other_alerts_snooze", "20"));
 
         long now = new Date().getTime();
+        
+        if (!bg_missed_alerts) {
+        	// we should not do anything in this case. Until there is a listener on UI changes, try again in 5 minutes
+            Log.d(TAG, "Setting timer to  5 minutes from now" );
+            checkBackAfterMissedTime(now + 5 * 60 *1000);
```
JohanDegraeve: the parameter of checkBackAfterMissedTime should actually be "in how much time should the check occur" , not "when"
If you call it with now + 5 \* 6 \* 1000, the alarm will actually be set to calendar.getTimeInMillis() + now + 5 \* 6 \* 1000

should be checkBackAfterMissedTime(5 \* 60 *1000);

tzachi-dar:
Thank you both.
I'll fix both comments tomorrow.

tzachi-dar:
ok, done what we have talked about, please review again.

AdrianLxM:
Looks good to me, thanks!

JohanDegraeve:
this looks not correct to me : 
long missedTime = bg_missed_minutes\* 1000 \* 60 - BgReading.getTimeSinceLastReading();
should be missedTime =  bg_missed_minutes\* 1000 \* 60 ;

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -268,6 +268,7 @@ public void onCreate(Bundle savedInstanceState) {
             addPreferencesFromResource(R.xml.pref_community_help);
 
             bindTTSListener();
+            bindBgMissedAlertsListener();
```
JohanDegraeve: be aware that this will bypass the check on alerts_disabled_until.
Actualy in AlertType a check is done on alerts_disabled_until , if this is higher than now, then you see that there's no call to checkIfMissedReadingAlert https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java#L104

If you now set this listener for preference change, a user may have all alerts disabled, but change the preference for missed readings, then that alert will not be suppressed anymore.

Normally get_highest_active_alert should be called every 5 minutes https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java#L102 so i don't think it's necessary that have that listener

JohanDegraeve: oops, not correct

tzachi-dar:
I think it is correct:
bg_missed_minutes is the time that after passes, we should alert. 
but if this time is 30 minutes, and we have BgReading.getTimeSinceLastReading() == 20 minutes, this means that we should alert in 10 more minutes.
And this is what the code is doing.

JohanDegraeve:
right, i seem to be not awake yet

tzachi-dar:
OK, fixed all issues that I saw and run some testing on it.
If one is ready to ignore the fact that we never start this intent at the first place, than I think that we are ready to go...
Comments are welcomed.

AdrianLxM:
Hm, what do you mean, we never start this service in the first place:

We start it, thanks to you, when the settings change.
We start it, when a new value comes in (via Notifications...).
Do we need to start it, when the app starts after a reboot?

In this case we could start the Service in `AutoStart.java`:

``` java
package com.eveningoutpost.dexdrip;

import android.content.BroadcastReceiver;
import android.content.Context;
import android.content.Intent;
import com.eveningoutpost.dexdrip.Models.UserError.Log;

import com.eveningoutpost.dexdrip.Services.MissedReadingService;
import com.eveningoutpost.dexdrip.UtilityModels.CollectionServiceStarter;

/**
 * Created by stephenblack on 11/3/14.
 */
public class AutoStart extends BroadcastReceiver {
    @Override
    public void onReceive(Context context, Intent intent) {
        Log.d("DexDrip", "Service auto starter, starting!");
        CollectionServiceStarter.newStart(context);
        context.startService(new Intent(context, MissedReadingService.class));
    }
}
```

tzachi-dar:
I believe we should start it on reboot and when xdrip starts. Otherwise it
will be funny if to start packets not arriving alerts we need to get a
packet.
בתאריך 20 בדצמ 2015 19:38,‏ "AdrianLxM" notifications@github.com כתב:

AdrianLxM:
On reboot yes, sure.
I believe the missed reading alert suffices in most cases (we don't start Notifications either, but I'm not sure about that). As soon as readings come in, we start Notifications-Service anyway. Well, there might be some corner cases with a snooze time almost over and no values comigng in... hm.

On start of the app (that is installation), do we really need to start it? It won't do no harm also starting it in xdrip.java as it is just once after installation.

tzachi-dar:
I think that we need this in xDrip.java it is used for after starting xDrip (for example after killing it from task manager, or after a crash).
It is true that other alerts don't start without a reading, but this is exactly the one that should always work in the case that there are no readings...
I'll try to prepare this soon.

tzachi-dar:
OK, I have fixed all issues I know with this commit, looking for any other comments.

AdrianLxM:
Another thing: xDrip (and the service timers) seem to be always running.
Are there cases where one would not use xDrip and thus want the timers to be stopped (e.g. no sensor session active)?

Or do we just say in this case, the user has to kill the app from the Android settings?

tzachi-dar:
Ideally, we would like to identify this situations and stop the timers.
Practically speaking, I believe that without a very good testing it is not feasible.
The code just gets more complicated and more complicated....

By the way, there is no way to kill xDrip from the settings. I have tried this and at least on my setup (I have a widget on the screen, don't know if this is the reason) xDrip will start again.
The good thing is that a timer every 5 minutes does not seem to cause any issues. (it takes battery, but one can live with that).

Specifically about this alert, looking at the code I'm afraid that it will fire even when the sensor is stopped. We can stop that as well but this makes the code more complicated. (this is probably needed, I'll open an issue about it).

I'm very scared of situation that timers will be fired all the time and this causes battery drain in ~90 minutes on my phone.
I also know that we have these bugs happening in another place (seen that at logs, but not quite sure where it is). So I think that we should be defensive and make sure that we don't fire the timers more than once in a minute .
This will at least stop the catastrophe of timers in the past.

AdrianLxM:
On KitKat it works killing xDrip. It doesn't work from "Running Applications" but from "Downloaded/Installed Applications".

tzachi-dar:
Merged this branch into master (I hope). Tried to do that by creating a new branch and rebasing it, but it seems that It did not do exactly what I wanted. But  the bottom line, this looks in master...

tzachi-dar:
As for stopping xDrip, from my experience on kitkat as well, it will be shown as stopped, but I see it running again after a few minutes...

AdrianLxM:
It depends on how I kill it. "Force Stop" seems to also kill all timers that could restart the app. It will also go through the whole startup (idempotent migrations, ...) when I click on the xDrip icon.

tzachi-dar:
Do you have a widget on the screen? I believe that this is what makes it start again...

AdrianLxM:
Yes I have a widget on the screen and on the lock-screen. They would simply not update. Did you try hitting both buttons "force stop" (and press ok that you can damage xDrip with that) and also "delete cache"?
I tested it again. It won't start again for me. If I stop it from the "running apps" it will just stop the services that will get restarted.
I also unpaired the wixel's bt chip first. Not sure if that was it?

jstevensog:
I often stop a sensor for a week, and I have never had issues with the timers.
I must have a go at making this indicate on the widget, as it maintains the last reading instead of showing any indication the sensor is stopped.  No biggie, because I stopped it, but it would be nice.
The PebbleTrend branch and watch face indicate the "no sensor" and display the traffic lights.  It would be nice if the widget at least displayed "--".

---- tzachi-dar wrote ----

tzachi-dar:
I believe that Adrian was refereeing to two issues:
1) having the timers around checks battery that is not needed...
2) closing the timers will make the xBridge not to receive acks, and as a result, he will have it battery drained...

PR #219, opened 2015-12-13
Make the MongoSendTask void, since no one uses it's return value.

tzachi-dar:


AdrianLxM:
Probably the best as we don't have an `onPostExecute()`.
You could also delete the

``` java
//        protected void onPostExecute(RSSFeed feed) {
//            // TODO: check this.exception
//            // TODO: do something with the feed
//        }
```

from the copied example, if it is still there.

thanks.

tzachi-dar:
I'll remove this in the next checkin.

Thanks
Tzachi

PR #220, opened 2015-12-14
Send queue improvments

tzachi-dar:
This PR fixes a lot of issues on send queue that we have talked about in the last month:

1) Use Upserts.
2) Correct use of wake lock.
3) Keep sending until there is nothing more to send (or error).
4) More efficient sending. (bigger number of objects being uploaded).
5) delete objects after using them in order not to make the db grow for ever.
6) upload more needed fields.
7) keep sending while there is work to do.

I have checked it with both mongo and NS and it seems that uploading 300 objects takes around 10-15 seconds on mongo and ~25 seconds on NS rest api.
On the other hand uploading 3 objects is much faster on the NS REST api (0.7 second compared to 4-5 seconds on mongo).
(There is a big fluctuation on measured times).

Checked this by uploading 500 objects every time. Worked well on mongo, but caused Azure to collapse (not sure it was realy this...).

I'll keep checking this with normal load for a few more days.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
@@ -56,18 +56,10 @@ public class BgSendQueue extends Model {
                 .orderBy("_ID desc")
                 .limit(nightWatchproMode ? 500 : 30)
                 .execute();
-    	if (!nightWatchproMode) {
-    		return values;
+    	if (nightWatchproMode) {
+    		 java.util.Collections.reverse(values);
     	}
-    	// swap the order of objects
-    	ArrayList<BgSendQueue> ret = new ArrayList<BgSendQueue>(values.size());	
-
-		int location = values.size() - 1;
-		for(BgSendQueue value : values) {
-			ret.set(location, value);
-			location--;
-		}
-		return ret;
+    	return values;
     	
     }
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
@@ -47,23 +47,13 @@ public class BgSendQueue extends Model {
     @Column(name = "operation_type")
     public String operation_type;
 
-/*
-    public static List<BgSendQueue> queue() {
-        return new Select()
-                .from(BgSendQueue.class)
-                .where("success = ?", false)
-                .orderBy("_ID asc")
-                .limit(20)
-                .execute();
-    }
-*/
     public static List<BgSendQueue> mongoQueue() {
         return new Select()
                 .from(BgSendQueue.class)
                 .where("mongo_success = ?", false)
                 .where("operation_type = ?", "create")
                 .orderBy("_ID asc")
-                .limit(3)
+                .limit(300)
                 .execute();
     }
 
@@ -83,7 +73,8 @@ public class BgSendQueue extends Model {
                 "sendQueue");
         wakeLock.acquire();
         try {
-            addToQueue(bgReading, operation_type);
+        	
+       		addToQueue(bgReading, operation_type);
 
             SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
 
@@ -172,10 +163,9 @@ public class BgSendQueue extends Model {
             wakeLock.release();
         }
     }
-
-    public void markMongoSuccess() {
-        mongo_success = true;
-        save();
+    
+    public void deleteThis() {
+        this.delete();
     }
 
     public static int getBatteryLevel(Context context) {
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
@@ -27,6 +27,7 @@ import com.eveningoutpost.dexdrip.ShareModels.BgUploader;
 import com.eveningoutpost.dexdrip.WidgetUpdateService;
 import com.eveningoutpost.dexdrip.xDripWidget;
 
+import java.util.ArrayList;
 import java.util.List;
 
 /**
@@ -47,14 +48,27 @@ public class BgSendQueue extends Model {
     @Column(name = "operation_type")
     public String operation_type;
 
-    public static List<BgSendQueue> mongoQueue() {
-        return new Select()
+    public static List<BgSendQueue> mongoQueue(boolean nightWatchproMode) {
+    	List<BgSendQueue> values = new Select()
                 .from(BgSendQueue.class)
                 .where("mongo_success = ?", false)
                 .where("operation_type = ?", "create")
-                .orderBy("_ID asc")
-                .limit(300)
+                .orderBy("_ID desc")
+                .limit(nightWatchproMode ? 500 : 30)
                 .execute();
+    	if (!nightWatchproMode) {
+    		return values;
+    	}
+    	// swap the order of objects
+    	ArrayList<BgSendQueue> ret = new ArrayList<BgSendQueue>(values.size());	
+
+		int location = values.size() - 1;
+		for(BgSendQueue value : values) {
+			ret.set(location, value);
+			location--;
+		}
+		return ret;
+    	
     }
 
     private static void addToQueue(BgReading bgReading, String operation_type) {
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
@@ -48,15 +48,15 @@ public class BgSendQueue extends Model {
     @Column(name = "operation_type")
     public String operation_type;
 
-    public static List<BgSendQueue> mongoQueue(boolean nightWatchproMode) {
+    public static List<BgSendQueue> mongoQueue(boolean xDripViewerMode) {
     	List<BgSendQueue> values = new Select()
                 .from(BgSendQueue.class)
                 .where("mongo_success = ?", false)
                 .where("operation_type = ?", "create")
                 .orderBy("_ID desc")
-                .limit(nightWatchproMode ? 500 : 30)
+                .limit(xDripViewerMode ? 500 : 30)
                 .execute();
-    	if (nightWatchproMode) {
+    	if (xDripViewerMode) {
     		 java.util.Collections.reverse(values);
     	}
     	return values;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
@@ -62,8 +62,8 @@ public class BgSendQueue extends Model {
                 .from(BgSendQueue.class)
                 .where("mongo_success = ?", false)
                 .where("operation_type = ?", "create")
-                .orderBy("_ID desc")
-                .limit(30)
+                .orderBy("_ID asc")
+                .limit(3)
                 .execute();
     }
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CalibrationSendQueue.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CalibrationSendQueue.java
@@ -41,8 +41,8 @@ public class CalibrationSendQueue extends Model {
         return new Select()
                 .from(CalibrationSendQueue.class)
                 .where("mongo_success = ?", false)
-                .orderBy("_ID desc")
-                .limit(20)
+                .orderBy("_ID asc")
+                .limit(2)
                 .execute();
     }
     public static void addToQueue(Calibration calibration, Context context) {
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CalibrationSendQueue.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CalibrationSendQueue.java
@@ -11,6 +11,7 @@ import com.activeandroid.annotation.Table;
 import com.activeandroid.query.Select;
 import com.eveningoutpost.dexdrip.Models.Calibration;
 
+import java.util.ArrayList;
 import java.util.List;
 
 /**
@@ -28,13 +29,26 @@ public class CalibrationSendQueue extends Model {
     @Column(name = "mongo_success", index = true)
     public boolean mongo_success;
 
-    public static List<CalibrationSendQueue> mongoQueue() {
-        return new Select()
+    public static List<CalibrationSendQueue> mongoQueue(boolean nightWatchproMode) {
+    	List<CalibrationSendQueue> values =  new Select()
                 .from(CalibrationSendQueue.class)
                 .where("mongo_success = ?", false)
-                .orderBy("_ID asc")
-                .limit(2)
+                .orderBy("_ID desc")
+                .limit(4)
                 .execute();
+        
+    	if (!nightWatchproMode) {
+    		return values;
+    	}
+    	// swap the order of objects
+    	ArrayList<CalibrationSendQueue> ret = new ArrayList<CalibrationSendQueue>(values.size());	
+
+		int location = values.size() - 1;
+		for(CalibrationSendQueue value : values) {
+			ret.set(location, value);
+			location--;
+		}
+		return ret;
     }
     public static void addToQueue(Calibration calibration, Context context) {
         CalibrationSendQueue calibrationSendQueue = new CalibrationSendQueue();
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CalibrationSendQueue.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CalibrationSendQueue.java
@@ -29,7 +29,7 @@ public class CalibrationSendQueue extends Model {
     @Column(name = "mongo_success", index = true)
     public boolean mongo_success;
 
-    public static List<CalibrationSendQueue> mongoQueue(boolean nightWatchproMode) {
+    public static List<CalibrationSendQueue> mongoQueue(boolean xDripViewerMode) {
     	List<CalibrationSendQueue> values =  new Select()
                 .from(CalibrationSendQueue.class)
                 .where("mongo_success = ?", false)
@@ -37,7 +37,7 @@ public class CalibrationSendQueue extends Model {
                 .limit(4)
                 .execute();
         
-    	if (nightWatchproMode) {
+    	if (xDripViewerMode) {
    		 	java.util.Collections.reverse(values);
     	}
     	return values;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CalibrationSendQueue.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CalibrationSendQueue.java
@@ -28,15 +28,6 @@ public class CalibrationSendQueue extends Model {
     @Column(name = "mongo_success", index = true)
     public boolean mongo_success;
 
-    /*
-    public static List<CalibrationSendQueue> queue() {
-        return new Select()
-                .from(CalibrationSendQueue.class)
-                .where("success = ?", false)
-                .orderBy("_ID asc")
-                .execute();
-    }
-    */
     public static List<CalibrationSendQueue> mongoQueue() {
         return new Select()
                 .from(CalibrationSendQueue.class)
@@ -52,9 +43,8 @@ public class CalibrationSendQueue extends Model {
         calibrationSendQueue.mongo_success = false;
         calibrationSendQueue.save();
     }
-
-    public void markMongoSuccess() {
-        mongo_success = true;
-        save();
+    
+    public void deleteThis() {
+        this.delete();
     }
 }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CalibrationSendQueue.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CalibrationSendQueue.java
@@ -37,18 +37,10 @@ public class CalibrationSendQueue extends Model {
                 .limit(4)
                 .execute();
         
-    	if (!nightWatchproMode) {
-    		return values;
+    	if (nightWatchproMode) {
+   		 	java.util.Collections.reverse(values);
     	}
-    	// swap the order of objects
-    	ArrayList<CalibrationSendQueue> ret = new ArrayList<CalibrationSendQueue>(values.size());	
-
-		int location = values.size() - 1;
-		for(CalibrationSendQueue value : values) {
-			ret.set(location, value);
-			location--;
-		}
-		return ret;
+    	return values;
     }
     public static void addToQueue(Calibration calibration, Context context) {
         CalibrationSendQueue calibrationSendQueue = new CalibrationSendQueue();
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoSendTask.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoSendTask.java
@@ -1,8 +1,10 @@
 package com.eveningoutpost.dexdrip.UtilityModels;
 
 import android.content.Context;
+import android.content.SharedPreferences;
 import android.os.AsyncTask;
 import android.os.PowerManager;
+import android.preference.PreferenceManager;
 
 import com.eveningoutpost.dexdrip.Models.UserError.Log;
 import com.eveningoutpost.dexdrip.Models.BgReading;
@@ -17,8 +19,6 @@ import java.util.List;
  */
 public class MongoSendTask extends AsyncTask<String, Void, Void> {
         private Context context;
-        //public List<BgSendQueue> bgsQueue = new ArrayList<BgSendQueue>();
-        public List<CalibrationSendQueue> calibrationsQueue = new ArrayList<CalibrationSendQueue>();
 
         PowerManager.WakeLock wakeLock;
         private static int lockCounter = 0;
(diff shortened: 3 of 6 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
```diff
@@ -47,23 +47,13 @@
...
-    }
-*/
     public static List<BgSendQueue> mongoQueue() {
         return new Select()
                 .from(BgSendQueue.class)
                 .where("mongo_success = ?", false)
                 .where("operation_type = ?", "create")
-                .orderBy("_ID desc")
-                .limit(30)
+                .orderBy("_ID asc")
```
AdrianLxM: This should be `.orderBy("_ID desc")` that new values get uploaded first as they are the more interesting regarding trends.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CalibrationSendQueue.java
```diff
@@ -28,21 +28,12 @@
...
-                .execute();
-    }
-    */
     public static List<CalibrationSendQueue> mongoQueue() {
         return new Select()
                 .from(CalibrationSendQueue.class)
                 .where("mongo_success = ?", false)
-                .orderBy("_ID desc")
-                .limit(20)
+                .orderBy("_ID asc")
```
AdrianLxM: Also here, should be `.orderBy("_ID desc")`

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoSendTask.java
```diff
@@ -16,19 +17,42 @@
...
+        private static int lockCounter = 0;
         private Exception exception;
         private static final String TAG = MongoSendTask.class.getSimpleName();
 
         public MongoSendTask(Context pContext) {
-            calibrationsQueue = CalibrationSendQueue.mongoQueue();
-            bgsQueue = BgSendQueue.mongoQueue();
             context = pContext;
+            PowerManager pm = (PowerManager) context.getSystemService(Context.POWER_SERVICE);
+            wakeLock = pm.newWakeLock(PowerManager.PARTIAL_WAKE_LOCK, "WifiReader");
```
AdrianLxM: I guess this should not be "WifiReader"?

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoSendTask.java
```diff
@@ -16,19 +17,42 @@
...
+            wakeLock.acquire();
+            lockCounter++;
+            Log.e(TAG,"MongosendTask - wakelock acquired " + lockCounter);
+
         }
 
         public Void doInBackground(String... urls) {
+            boolean sendMore = true;
+            try {
+                while (sendMore) {
```
AdrianLxM: This might lead to aggressively uploading data and block the whole internet connection and may lead to additional costs if someone goes online out of the house just to check a website.
The service that starts this task will be called every few minutes anyway so there is no need to continue.

A function that can upload month of data at once should have to be triggered by the user (Button in the menu or settings e.g.)

At leas we should check that this does not happen on mobile data but just on Wifi.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -209,6 +211,11 @@ private void populateV1APIBGEntry(JSONArray array, BgReading record) throws Exce
             json.put("unfiltered", record.usedRaw() * 1000);
             json.put("rssi", 100);
             json.put("noise", record.noiseValue());
+            json.put("xDrip_raw", record.raw_data);
+            json.put("xDrip_filtered", record.filtered_data);
+            json.put("xDrip_calculated_value", record.calculated_value);
+            json.put("xDrip_age_adjusted_raw_value", record.age_adjusted_raw_value);
+            json.put("sysTime", format.format(record.timestamp));
```
AdrianLxM, line 221: Why do we need this uploaded?
Especially as most of it is duplicates.

This will blow up network use and make NightScout more expensive.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -233,6 +240,13 @@ private void populateV1APIMeterReadingEntry(JSONArray array, Calibration record)
             json.put("date", record.timestamp);
             json.put("dateString", format.format(record.timestamp));
             json.put("mbg", record.bg);
+            json.put("xDrip_slope", record.slope);
+            json.put("xDrip_intercept", record.intercept);
+            json.put("xDrip_estimate_raw_at_time_of_calibration", record.estimate_raw_at_time_of_calibration);
+            json.put("xDrip_slope_confidence", record.slope_confidence);
+            json.put("xDrip_sensor_confidence", record.sensor_confidence);
+            json.put("xDrip_raw_timestamp", record.raw_timestamp);
+            json.put("sysTime", format.format(record.timestamp));
```
AdrianLxM, line 254: Did you talk to the NightScout people. Will they actually use this data on the website?
Also the data does not look like a meter reading but like calibration data.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -253,6 +267,7 @@ private void populateV1APICalibrationEntry(JSONArray array, Calibration record)
                 json.put("intercept", (long) ((record.intercept * -1000) / (record.slope * 1000)));
                 json.put("scale", 1);
             }
+            json.put("sysTime", format.format(record.timestamp));
```
AdrianLxM, line 275: this is a duplicate we already upload. Did you talk to Jason about that?

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
tzachi-dar: The problem with uploading them in reverse order, is that it makes it very hard to close gaps (while remaining network efficient).
Now that uploading is much faster, I believe that it does not really matter.
For example closing a gap of a day, will  now take ~15 seconds.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoSendTask.java
tzachi-dar: yes, that is correct, I'll fix that soon.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoSendTask.java
tzachi-dar: OK, I'll change that so there will not be a very long loop here.

I would still like to call it faster than to wait for the next 5 minute though.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
tzachi-dar, line 275: yes, this was discussed with Jason on gitter on first of december.

The main reason that this is need is since this the field that NS are using to not have duplicates.
If this is not used, then for each calibration/meter record we will have two records representing the one measurement that we have done.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
tzachi-dar, line 254: The data that starts with xDrip is the data that is needed for xDrip to work correctly, in more details, this are needed for the calibration graph and and calibration table.

As for putting this here or on the calibration. Here are my thoughts about this:
We would like the NightwatchPro to work with NS data and also with the xDrip mode. Obviously people that work with xDrip will have more data (like calibration graph and so on).

NS has divided the data that xDrip calls calibration into two places, one is a meter record and one is the calibration itself. Both by the way, don't contain the exact fields that xDrip uses for calibration. This is why I need extra fields.
So, why did I put it here? well the most important data in a calibration (that is displaying it on the screen) is the time and the actual value measured. 
So, reading this "meter record" will allow me to display the data for xDrip users as well as for NS users.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoSendTask.java
AdrianLxM: In this case we would have to reduce it to 30 uploads again to not have such a data load at once.
I prefer the system not to wake up that often to safe battery. That's why @mgranberry increased the intervals.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
AdrianLxM: The most recent datum must always be uploaded first when it comes in.
If not, backfilling data from 2 month ago will stop people from receiving new values.

tzachi-dar: These depends on how fast you upload the data.
As I have said, I can upload 300 packets (or a day data in 30 seconds).
So, in the very rare case that someone did not upload his data for 2 months he will have to wait for 30 minutes.
In most normal cases the delay will only be of a few seconds.

tzachi-dar: If this helps, I can start by uploading data from the last 2 days, and only than, all other data.
In that case delay will be less than a minute.

AdrianLxM:
I'm fine with a fast backfill of the last one or two days and then upload a day every "6" minutes.

If I'm not mistaken having the timer set at 6 minutes will in the standard case just push the timer further and the uploads happens every 5 minutes when a new value comes in.

tzachi-dar:
OK, so I'll upload the last 2 days and then every 6 minutes anything that is older than that.
I believe that practically, this will happen with every new packet.

AdrianLxM:
I'm still not sure what the purpose of the `xDrip-...` fields are. Do you...

1) ... just use it to remotely have the calibration graph and calibration data table?
2) ... really sync the internal data structures to later be able to hand over the CGM session to another phone?

The handover (2)) would be a nice feature to easily swap phones ("the small one for activity, the big one during the work day" or "to charge the phone and have the kid run around with another one")... but it would make it hard to handle data integrity as calibrations could now be made with more than one phone at the same time.

Just for having the calibration graph possibly shown remotely (1)) it is a lot of data that gets uploaded and downloaded more for a very small feature. 

In any case, this data would be uploaded even for people not using this feature (not need it, or just use NightScout) and downloaded again for viewers that cannot process the data anyways (like programmes/widgets for iPhone, the Chrome-Plugin, ...).

To solve this, I would propose the following:
- Just upload the xDrip-internal data when the user activates that in the settings.
- Do not upload it together with the common datatypes `cal`, `sgv`, ... but introduce new types like `xdrip-internal_cal`, `xdrip-internal_BgReading`, `xdrp-internal_sensor`, ... 
  This way you can still filter for those at the `entries.json` endpoint of the NightScout-API (`/api/v1/entries.json?find[type][$eq]=xdrip-internal_cal`). But other programmes that just filter the standard sgv don't have more load.

cheers, Adrian

AdrianLxM:
> OK, so I'll upload the last 2 days and then every 6 minutes anything that is older than that.
> I believe that practically, this will happen with every new packet.

On mobile data direct upload to mongo might still take much much more time than 15 seconds for one day of data.
Also the upload can fail after the first couple of values already being uploaded. Not reverse-ordering the data might lead to never showing the current data on NightScout (until a wifi-connection is established or you enter in an area with good 3G coverage).
Not reverse ordering the values therefore will need some logic to ensure, that once the upload fails, the current value(s) will then get uploaded first - without retrying instantly to do so after a failed upload as the line might still be busy.

tzachi-dar:
Indeed allowing one to switch xDrips would be nice, but it seems to me too much complicated at this stage.

As for calibration, this really has no influence on anything. People calibrate twice per day or 5 times per day. The real issue (if there is an issue) is what about BG data.

As for the issue of data size. I have captured a reading of one entry from the nightscout. Attached is the data that was read:
it consists of 18 packets with a total size of 6631 bytes.
Obviously, most of the bw is spent on overhead of opening the connection, authenticating, and so on.
if the data size if 20 bytes or 30 bytes it has no real effect on what the user will pay for bw, or the time that it will take.
Attached is the file with data: 

[reading_from_ns.pcapng.zip](https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/files/61376/reading_from_ns.pcapng.zip)

so, do you still fill that bw is a problem in this scenarios?

tzachi-dar:
After some more thinking, will a setting saying "upload data for NightwatchPro" solve both problems?

This setting will allow control:
1) the new fields to upload.
2) the order to send them.

Thanks
Tzachi

AdrianLxM:
It is a different kind of data and should be encapsulated in different kind of type. Especially if the datastructures already are designed like this.
It is (or will be) just an exchage format between two xDrip instances, not affecting any other programme in the NightScout ecosystem.
Other data like treatments weren't added to the existing data structures but uploaded as different type or even in a different database document.

I would not call the setting "upload data for NightwatchPro" but "Upload internal xDrip datastructure".

JohanDegraeve:
allowing switch xDrips would indeed be very nice, it could even be automatic. If xdrip looses connection with device A, device B which is around could connect to it. If both devices keep synchronized using nightscout as cloud.

In fact this idea is on my mind since the beginning I learned about xDrip (which is two months ago).
I did a similar thing with HelpDiabetes using Google Fusion Tables as cloud. One can add/modify/delete objects (meal contents, bg meter results, insulin events, all settings), everything is well synchronized between all devices. (i would like to add the apk here but apparently github doesn't accept apk as attachment)
So it's well doable.
What you need : a unique id per object (unix timestamp is most suited, already suggested here), probably also a lastmodifiedtimestamp because I assume some objects could also get updated (eg in case of override calibration ?)

As far as bandwidth is concerned : 99 % of the cases a device will query nightscout for the objects that got created/updated less han 5 minutes ago, that should be rather small ?

tzachi-dar:
Switching between xDrips can indeed be interesting, although I must say that I'm trying to achieve most of that with the following changes, that I'm making.

My main goal, is to have a master xDrip that has all data locally but allow other ones to display everything that is interesting.
Indeed one has to keep 3 tables of xDrip synced (sensors, bgreading and calibrations).
One of them will be a master, and the others will be slaves. 
Keeping the data in sync indeed require a query every 5 minutes, that can be achievable.
The main problem with this kind of projects is how to deal with corner case like having the network disconnected, duplicate data, master selection and so on.
Can be an interesting challenge, but I think that if we indeed want to reach there, we need to go slowly.

JohanDegraeve:
"My main goal, is to have a master xDrip that has all data locally but allow other ones to display everything that is interesting."
==> that would indeed already be very good.

"The main problem with this kind of projects is how to deal with corner case like having the network disconnected, duplicate data, master selection and so on."
==> as soon as network is back, the new query would be for a longer time period, ie all objects created/updated since last time there was a connection or last time sync was successfully finished (the oldest).
Avoid duplicates using the unique id and compare. If there's two updates of one object, then the one with the highest creation/modified timestamp is the 'master value' - it's the uploading device that should assign this timestamp (not the nightscout api as it is now I believe).
There's no master device, just a most recent version of each object.

unique id should be timestamp in milliseconds (not seconds)

JohanDegraeve:
if it's indeed the aim also to update objects, (calibrations ? sensor ?) then the nightscout api should allow updates, i've been searching yesterday evening , it looks like it only allows creation and delete, not updates
i'll test that further this evening

AdrianLxM:
@JohanDegraeve, this update adds "sysTime". With adding `sysTime` it will automaticalyll be an upsert with "sysTime" and "type" as query.

"sysTime" can basically anything. A timestamp, a string, an ID.

If we had our own types (not interfering with the NightScout types), we actually could put an ID, not a timestamp.

tzachi-dar:
OK, let's examine more carefully how xDrip works:

There is a list of sensors. each sensor has calibrations. Each calibration has bg readings.
Each calibration has a slope and intercept that are probably the most important fields in it.
A slope and intercept are calculated based on data from the last 4 days. See the function (calculate_w_l_s() for the dirty details (challenge question, it has a bug find it.)

This is the base. now let's consider the two xDrip not working together. now, each can have its set of calibrations,  one adds a calibration to the first xDrip. while the parent near the kid replaces a sensor. now both will calculate different BG values that are based on different calibrations/sensors. now the later will have a calibration added. now, the later has more updated data, but on some place in the middle the other had fresher data. (all this data will have the same timestamp, it's the calibrations / sensors that matters). So deciding which object to keep is more than just a timestamp problem...

But I guess that the most problematic thing to solve here is not this theoretical problems but the fact that xDrip was not designed for this and has a lot of places where the data gets changed.

As for the technical details of updating a record. Well this is being done using a field that is called sysTime.

you can see that in this very specific checkin, I'm adding it both to the mongo api and to the rest api.
So once I'll fix what's need fixing here and commit this, we will have updates possible. We will also not have 2 copies of the same data ...

By the way as for timestamps, since we can not get the clocks synced, I don't think that you can trust a difference of 10 seconds, when telling who was first. (we do use ms but...)

tzachi-dar:
The main reason that I want to add fields to this existing objects and not to create new objects is to allow nightwatch pro to work with NS.
For people that don't have xDrip will show the data that we can.
People with xDrip will enjoy the benefits from more data, but all people can enjoy the benefits of all the fixes that we are doing to xDrip.
The mongo architecture allows one to add more fields, and applications that don't use this fields will not be hurt.

JohanDegraeve:
@tzachi-dar , you say "This is the base. now let's consider the two xDrip not working together"
this is not how I interpreted this discussion. Indeed that's indeed something totally different and more difficult. But I had the impression you had already been working on that , like the subject about letting xdrip and wifi wixel work in parallel ?

regarding switching xdrips : I was more thinking of one single xdrip but more than one Android device runninng the xdrip Android app. As you were mentioning "My main goal, is to have a master xDrip that has all data locally but allow other ones to display everything that is interesting"
That I think should be feasible , a lot of work but doable. 
Basically all data that is stored in tables, could be kept in sync via the cloud. Let's say xdrip android app (to be clear) number 1 has the bluetooth connection to the xdrip and uploads measurements to the cloud. In addition to calibration data.
xdrip android app number 2 doesn't have the bluetooth connection but it's fully aligned : it will regularly check in the cloud if there's any data changed since the last sync (xdrip measurements, calibrations, ...) and if so downloads them and stores them locally. As a result xdrip2 will display exactly the same values (same algorithm with same data as input). 
Let's assume xdrip android app 1 looses the bluetooth connection, but app 2 is close to the xdrip, it will connect (assuming it had the connection before so it knows the bt device), and just continue doing the job that app 1 was doing : receive measurements, and upload them to the cloud ...

Result : nightwatch becomes fully useless. Just one app with all functionality. Remote users with the app running can follow, have the alarms etc.

But I'm still not sure if the nightscout data and api would be sufficient at the moment to support this.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
mgranberry: Nightscout should handle batches of 10k records.  I use 5k in my uploader to be conservative.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoSendTask.java
mgranberry: If there is data waiting there is nothing wrong with sending it, but the uploader should probably register for network state/connectivity broadcasts and perform big batch backfills only while on unmetered connections, preferably while plugged in.  It should also implement exponential backoff to avoid excessive retries, especially if the reattempts are very frequent.  I intentionally left it gated on new readings being available because the mobile radio will be woken up already, so extra data isn't very expensive power-wise.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -233,6 +240,13 @@ private void populateV1APIMeterReadingEntry(JSONArray array, Calibration record)
             json.put("date", record.timestamp);
             json.put("dateString", format.format(record.timestamp));
             json.put("mbg", record.bg);
+            json.put("xDrip_slope", record.slope);
```
mgranberry: This is cal entry data.  Is the slope/intercept going to change much?  I think these could reasonably be put there and sent only when new calibrations happen.  Additional data won't be very "expensive" there.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
mgranberry, line 254: Would it be possible to include a "calibrationId" field that points to the cal entry?  That would be much less overhead for an edge case.  The _id field can be set by the uploader for nightscout uploads, so xDrip can be sure about the entry being present.

mgranberry, line 254: It is also possible to add additional entries, e.g. xDripCal, that contain xDrip-specific data.  If that's done, please hide it behind a flag.  This looks like a lot of noise and I'm already burning through more data than I would like on my watch/phone.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoSendTask.java
tzachi-dar: I'll change that to return to send only when there is a new reading coming.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
tzachi-dar, line 254: There is a timestamp that can be used as a calibration id.
There is no plan to show more data than we already have today.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
tzachi-dar: this is exactly the thing that will happen. only sent when there is a new calibration, or change in existing one.

tzachi-dar:
@JohanDegraeve in your suggestion, you are talking about the good flow of data. This is somewhat hard but possible.
What you are missing here is the bad flows:
For example:
Consider two devices, holding the tables. One of them or both looses connection to the cloud. now their tables are different... Now the question is how to sync them again. I don't think it is a trivial question.
Another issue: 2 xDrip devices, one from version a, another from version a+1. How will the new one work with the fields of the new one that he is not aware of.
Because of this kind of problems I don't see how your suggestion will become feasible.

# 

That said, I believe that even today we are very close to what you suggest:

You have one android device running xDrip. It is reading data from BS and from the cloud. it is near the person with transmitter, and all is well. (this is device a)
Next stage, one using device b, going with the kid. device b is running dexterity (another android application). dexterity connects with the BT device and uploads data to the cloud. 
Device a reads the data, and continues to be updated. it is generating BG values.
Device B is receiving the data and showing it like any android device.

All this is working right now (dexterity needs to be upgraded for the big crowd).
So, what are you missing in this architecture?

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
AdrianLxM: This is the first time I see an advantage of setting up NightScout for adults using mongo just as data backup ;)
With mongo-uploads I guess it is more a weak gsm connection that screws up things than mongolab. But a few days a go when we talked about upping the amount of data, I also upped the number and lowered the repitition time. After some uploads it always ended in failed uploads for a short period of time.
It might have been only the upload of one value that actually failed, but it always causes the whole batch to keep being marked not uploaded. :/
As long as these cases can happen, I'm strongly in favour of a solution that sets priority to the most recent "live" value to be uploaded. (Even though I personally just upload once in a while and on wifi and would personally profit from a faster burst upload.)

tzachi-dar: If one is only uploading the data once in a month, how comes the live
results are suddenly that important?

But in any case, as we have said, uploading order will depend if you need
the data for nightscout pro or not.

AdrianLxM: > If one is only uploading the data once in a month, how comes the live results are suddenly that important?

In such cases where adults need remote-monitoring (like severe sickness and the spouse cannot stay at home, hospital stays after accidents...) there usually is not much time and ease to do long syncs and setups.

tzachi-dar: I believe that in the case of adults taking care of a child, data is
constantly uploaded.

But in any case, for people who don't want to have nightwatchpro, and did
not change that checkbox nothing will change.

JohanDegraeve:
@tzachi-dar ok to me, i need to try out dexterity

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
JohanDegraeve: as adult I use nightscout. Being 10 hours per day behind a PC, i just need to hover over the IE tab in the taskbar (windows 7) to see the value and the direction.
Pebble would be an alternative but I never used watches, I don't think I would wear it.

tzachi-dar:
you can find it out here:
https://github.com/tzachi-dar/dexterity.git

probably the first thing is to make sure that it can connect to a bridge
that is not hard coded.
Please note it does not yet support the xBridge protocol, only xDrip.

tzachi-dar:
Did most changes, expect for the UI itself. all comments are welcomed.

AdrianLxM:
Hi @tzachi-dar, Thanks for all the work! I didn't have the time to look through all. Just one note:
In order to reverse a Collection, there is a built in function in the Java API:
`java.util.Collections.reverse(values);`

AdrianLxM:
Hi @tzachi-dar, I like the changes you have made.

What I somehow disagree is the naming of "NightWatchPro" as it isn't a successor of NightWatch but an extended Version of xDrip. How about "xDripViewer"? Or simply "xDrip 3"?

tzachi-dar:
xDripViewer can be great, although I must say I want it to read any nightscout data. But if people don't have xDrip, than they will have less data, but they can still see many things.

tzachi-dar:
OK, changing that to java.util.Collections.reverse.

AdrianLxM:
If we have integrated everything in xDrip it would be just xDrip - not another programme.

tzachi-dar:
Just did some last cosmetic changes to this branch.

tzachi-dar:
This was merged into master as a new pull request.

PR #222, opened 2015-12-15
Make sure that in the case that no good calibration exists, reasonable…

tzachi-dar:
… data will be used for slope and intercept.

Seems that if a calibration exists but with slope_confidence == 0
the function allForSensorInLastFourDays() might return with a list with size 0.
This will lead it to reaching the line 
 double d = (l \* n) - (m \* m);
and d will be 0.
As a result, 
calibration.intercept = ((n \* p) - (m \* q)) / d;
will be divided by 0, which means bad things later...

By mistake, pushed it already at master, but please review and let me know if this seems a good fix.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
@@ -393,7 +393,7 @@ public class Calibration extends Model {
                 .execute();
     }
 
-    public static void calculate_w_l_s() {
+    private static void calculate_w_l_s() {
         if (Sensor.isActive()) {
             double l = 0;
             double m = 0;
@@ -407,6 +407,7 @@ public class Calibration extends Model {
                 calibration.slope = 1;
                 calibration.intercept = calibration.bg - (calibration.raw_value * calibration.slope);
                 calibration.save();
+                CalibrationRequest.createOffset(calibration.bg, 25);
             } else {
                 for (Calibration calibration : calibrations) {
                     w = calibration.calculateWeight();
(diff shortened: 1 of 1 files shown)


AdrianLxM:
**I'm good with this change, but would think of adding the following line:**
`CalibrationRequest.createOffset(calibration.bg, 25);`
and setting `calculate_w_L_s()` to `private`.

+++++++++++
You may stop reading here ;)
+++++++++++

My reasoning: (More a documentation for myself)
- `Calibration.last()` will only give us Calibrations for this sensor.
  -> The logic now is "If there is *\* _no or_ *\* just one calibration within the last four days, use 1 as slope and calculate the intercept from the last known (bg, raw) pair."
- `calculate_w_L_s()` is always just called after a `safe()`. So we can be sure there is a `Calibration.last()`. (Before we could be sure as there was one within the last four days). -> set `calculate_w_L_s()` to private
- with zero values, we cannot do linear regression.

**But: Running into the error case would trigger the slopeOOB-handler that may:**
- return the slope of an earlier calibration that is not marked as "bad". (Not sure if this is good anyways because of *)
- use `estimate_raw_at_time_of_calibration` instead of `raw_value` to calculate the intercept (so it may be the age-adjusted raw value). This may be a good thing but then again if we use the other one if we have one more calibration.
- **_Register a request for a new calibration**_ (I have the alarms for them disabled anyways ;) )

**|\* We can only have this situation when:|**
- There was "future data" deleted (This leaves us in a bad state anyways: It can give us bad or overwritten calibrations as last one -> user has to enter a new calibration anyways)
- There was no calibration entry within the last 4 days
- There were only calibration entries where the 5-minute-delta was >= 16 mg/dl or... (slope_confidence == 0)
- the entered value was below -0.55 or above 370.385 mg/dl (sensor_confidence == 0)

tzachi-dar:
OK, updated accordingly, please review.

AdrianLxM:
Yea, looks good to me! (Hard to life test though).

tzachi-dar:
not, if you have a develop system...
change the 4 days to 1 hour, the slope confidence to 0.1 and add a calibration.
Without this change you will crash immediately...

AdrianLxM:
My test phone already was delivered to the house of my parents... I will pick them up when we visit them for Christmas.
In my opinion you can merge it.

Issue #180, opened 2015-11-22
No local broadcast with Share


AdrianLxM:
Bug report via chat with @KeesaKat / Marybeth: 
1. xDrip doesn't broadcast to NightWatch (and HAPP), if the value is not calculated ("???").
   That would be good to have as NightWatch in recent versions can show raw values as well.
2. The local broadcast does not broadcast raw but copies the value if  used with a share receiver.

Entropy512:
OK, number 2 explains what I've been seeing since moving to the latest NightWatch, I can confirm that the raw values are always identical to the "filtered" value

AdrianLxM:
@KeesaKat, @Entropy512: This PR should solve the issue: https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/223

Would you like to test it?

AdrianLxM:
#223 was also a fix for this issue.

PR #223, opened 2015-12-17
slope and intercept transformation for NightScout

AdrianLxM:
**1) Calculation of Slope and Intercept for NightScout**

In xDrip(wixel) we calculate the values in the form: `raw*slope + intercept`
NS (and dexcom?) calculate the values in the form: `(raw - intercept)/slope`

Also raw values are 1000 times of what we have in xDrip.

You can rearrange the xDrip formula like this:

`s*r + i = (1000*r - (i * -1000/s) ) / (1000/s)`
(Test this with wolfram-alpha: http://www.wolframalpha.com/share/clip?f=d41d8cd98f00b204e9800998ecf8427eh82hpg790s)

This is the same as `(NS_raw - NS_intercept) / NS_slope` if:
- NS_raw = 1000*r
- NS_intercept = (i \* -1000/s)
- NS_slope = (1000/s)

**2) Upload Slope and Intercept with higher precision**
NightScout can handle float values. -> higher precision

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java
@@ -120,12 +120,12 @@ public class BgSendQueue extends Model {
                         intercept= cal.first_intercept;
                         scale =  cal.first_scale;
                     } else {
-                        slope = cal.slope * 1000;
-                        intercept=  (cal.intercept * -1000) / (cal.slope * 1000);
+                        slope = 1000/cal.slope;
+                        intercept=  (cal.intercept * -1000) / (cal.slope);
                         scale = 1;
                     }
-                    unfiltered= bgReading.usedRaw();
-                    filtered = bgReading.ageAdjustedFiltered();
+                    unfiltered= bgReading.usedRaw()*1000;
+                    filtered = bgReading.ageAdjustedFiltered()*1000;
                 }
                 //raw logic from https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/plugins/rawbg.js#L59
                 if (slope != 0 && intercept != 0 && scale != 0) {
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
@@ -237,6 +237,10 @@ public class NightscoutUploader {
         }
 
         private void populateV1APICalibrationEntry(JSONArray array, Calibration record) throws Exception {
+
+            //do not upload undefined slopes
+            if(record.slope == 0d) return;
+
             JSONObject json = new JSONObject();
             SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ", Locale.US);
             format.setTimeZone(TimeZone.getDefault());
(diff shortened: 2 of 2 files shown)


stavlor:
:+1:

tzachi-dar:
As for the changes in the file nightscout uploader they seem to be right. 

from the mathematical standpoint, results of what happens when cal.check_in == true (for example share) change by a factor of 1000.
Did we have such a bug or this is something that you have forgot on your checkin?

Also, it seems that even for the normal case your calculation changes things. I guess that this was the intention.

In any case, I would write that like this (at least make the xdrip native code simple): 
if (cal != null){

// slope/intercept/scale like uploaded to NightScout (NightScoutUploader.java)
unfiltered= bgReading.usedRaw();
filtered = bgReading.ageAdjustedFiltered();

```
                if(cal.check_in) {
                    slope = cal.first_slope;
                    intercept= cal.first_intercept;
                    scale =  cal.first_scale;

                   if (slope != 0 && intercept != 0 && scale != 0) {
                       if (filtered == 0 || bgReading.calculated_value < 40) {
                          raw = scale * (unfiltered - intercept) / slope;
                        } else {
                       double ratio = scale * (filtered - intercept) / slope / bgReading.calculated_value;
                       raw = scale * (unfiltered - intercept) / slope / ratio;
                     }

                } else {
                    slope = cal.slope;
                    intercept=  cal.intercept;

           //raw logic from https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/plugins/rawbg.js#L59
            if (slope != 0 && intercept != 0 && scale != 0) {
                if (filtered == 0 || bgReading.calculated_value < 40) {
                    raw = unfiltered * slope  + intercept
                } else {
                    double ratio = (filtered * slope + intercept)  / bgReading.calculated_value;
                    raw = unfiltered * slope  + intercept / ratio;
                }
            }

                }

            }
```

This has the disadvantage of duplicating the if code, but it is defiantly better in the mean that xDrip calculations remain xDrip calculations and there is no need to convert the parameters from xdrip to NS and than calculate them like NS. xDrip should use the same method in all places.

also for the ns code, I don't see a real value in multiplying every variable in 1000 and then dividing them in 1000...

AdrianLxM:
- Multiplying by 1000. NightScout us accepts integer values as raw data. Multiplying by 1000 is like giving it 3 digits after the decimal point. So being larger high by a factor of 1000 is the internal representation of raw data from Dexcom and NightScout. So any changes pleas point there ;)
- The _checkin_ calibrations already have the other dimension as they come from a Dexcom receiver.
  - You can see a cal record posted by @stavlor on gitter: `{"_id":"5672c095b26a80e8a6bb07b3","device":"xDrip-DexcomShare","type":"cal","date":1450156415019,"dateString":"2015-12-15T00:13:35.019-0500","slope":672,"intercept":28532,"scale":1}`
  - This is how we do the calculation in xDrip as well for "checkins". The representation already is like in NightScout and we convert it back:

``` java
            if(calibration.check_in) {
                double firstAdjSlope = calibration.first_slope + (calibration.first_decay * (Math.ceil(new Date().getTime() - calibration.timestamp)/(1000 * 60 * 10)));
                double calSlope = (calibration.first_scale / firstAdjSlope)*1000;
                double calIntercept = ((calibration.first_scale * calibration.first_intercept) / firstAdjSlope)*-1;
                bgReading.calculated_value = (((calSlope * bgReading.raw_data) + calIntercept) - 5);
```
- As for the Formulat used in BgSendQueue: The "raw" shown in NightScout is a different thing than we understand in xDrip. It may have a predictive part as well. The value is not used inside xDrip but broadcast to NightWatch (and HAPP, ...) to be able to show the same "raw" representation that would come from the NS server. This is why the formula is the same as in https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/plugins/rawbg.js#L59 and the input is the same as we upload.

AdrianLxM:
By filtered and unfilterd being multiplied by 1000 they now should be good for Share-parameters and it should also solve the raw issues with Share: https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/issues/180

AdrianLxM:
@tzachi-dar: My idea is to refactor the functionality into the following methods:

Calibration.getNightScoutScale()
Calibration.getNightScoutSlope()
Calibration.getNightScoutIntercept()
BgReading.getNightScoutRaw()

Then we wouldn't have to make the distinction between _checkin_ or not in the NightScout-Uploader as well. And we can call the function in the Mongo upload, the NS-Rest upload and for the local broadcast.
This PR is just a fix. I will do the refactoring once your version of NightScout-Uploader is in.

tzachi-dar:
Thanks for your explanations. Seems quite a mess, but probably started before this checkin...

PR #227, opened 2015-12-22
Alerts times

tzachi-dar:
First draft of the way I think this should be fixed, if we decide to take the hard way...


AdrianLxM:
Hi @tzachi-dar, it seems like you pushed a lot of other stuff you are working on.

tzachi-dar:
you are right opening a new one.

PR #228, opened 2015-12-23
Alerts times3

tzachi-dar:
This is the fix code.
It still needs work and testing.
Any comments are welcomed.

Please note that I'll not be available next week, and therefore would like to have things commited in a day or two. 
I believe that it makes things much better even when not finished...

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
@@ -295,29 +295,73 @@ public class Notifications extends IntentService {
         }
     }
 
-    private void ArmTimer(Context ctx) {
-        Log.d(TAG, "ArmTimer called");
-        ActiveBgAlert activeBgAlert = ActiveBgAlert.getOnly();
-        if (activeBgAlert != null) {
-            AlertType alert = AlertType.get_alert(activeBgAlert.alert_uuid);
-            if (alert != null) {
-                Calendar calendar = Calendar.getInstance();
-                AlarmManager alarm = (AlarmManager) getSystemService(ALARM_SERVICE);
-                // sleep longer if the alert is snoozed.
-                long wakeTime = activeBgAlert.next_alert_at;
-                Log.d(TAG , "ArmTimer waking at: "+ new Date(wakeTime) +" in " +  (wakeTime - calendar.getTimeInMillis())/60000d + " minutes");
-                if (wakeIntent != null)
-                    alarm.cancel(wakeIntent);
-                wakeIntent = PendingIntent.getService(this, 0, new Intent(this, this.getClass()), 0);
-                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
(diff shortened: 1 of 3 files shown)


AdrianLxM:
Wouldn't this disable the functionality of clearing the active alert when disable alert is pressed?
What is the plan to have this functionality then? An extra button?

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
```diff
@@ -295,29 +295,73 @@ public void notificationSetter(Context context) {
         }
     }
 
+    private long CalcuatleArmTime(Context ctx, long now) {
```
AdrianLxM, line 298: The Java convention has only constructors starting with a capital letter.

tzachi-dar:
I believe that this will disable the functionality, but two questions:

1) why do we need this functionality in first place?
2) I believe that the functionality is to snooze the alerts not to delete them all together. When they are deleted, we can't look at things like: activeBgAlert.next_alert_at

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -470,20 +471,25 @@ public void onClick(DialogInterface dialog, int which) {
...
 
         
         private void bindBgMissedAlertsListener(){
-          findPreference("bg_missed_alerts").setOnPreferenceChangeListener(sBgMissedAlertsHandler);
-          findPreference("bg_missed_minutes").setOnPreferenceChangeListener(sBgMissedAlertsHandler);
-          findPreference("other_alerts_snooze").setOnPreferenceChangeListener(sBgMissedAlertsHandler);
+          findPreference("bg_missed_alerts").setOnPreferenceChangeListener(sAlertsPreferanceHandler);
+          findPreference("bg_missed_minutes").setOnPreferenceChangeListener(sAlertsPreferanceHandler);
+          findPreference("other_alerts_snooze").setOnPreferenceChangeListener(sAlertsPreferanceHandler);
+          findPreference("alerts_disabled_until").setOnPreferenceChangeListener(sAlertsPreferanceHandler);
```
AdrianLxM, line 490: This "setting" isn't changed by the user and not in this activity.
It will be changed programmatically after a button press in the SnoozeActivity.

AdrianLxM:
As a programmer and tester I need the functionality to cancel active alerts and have a fresh alert on the next incoming value.

As a user I need it as the alarms don't cancel themselves when the value is in range again before the snooze time is over.

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
tzachi-dar, line 490: I understand, but if the handler is called, than I think it is better to have all the callbacks in one place. (did not try that yet).

tzachi-dar:
Since we are having the timer called every time one of the snoozes will be over, I believe we can try and return that code and see how this will work.

tzachi-dar:
closing this, since it needed many things undone.

PR #231, opened 2015-12-26
Navdrawer with higher contrast

AdrianLxM:
Black text on white background


tzachi-dar:
Can't look at this right now, but if you can fix the colors for the debug
strings this will be great...
בתאריך 26 בדצמ 2015 18:15,‏ "AdrianLxM" notifications@github.com כתב:

AdrianLxM:
Good Idea, @tzachi-dar, these were even harder to read.
I set the text colour to black also.

JohanDegraeve:
much better.
but what is listitem_navigation_drawer.xml for ?

AdrianLxM:
>  but what is listitem_navigation_drawer.xml for ?
> It is the layout for the text shown in the items of the ListView of the Navigation Drawer.
> The built in 'android.R.layout.simple_list_item_activated_1' we used before will have the general text colour we use in the app which is while.
> It is basically the same layout as the built in one (for current built version); just the colour is different.

I didn't find any indication that something like "layout inheritance" (xml, not talking about inheritance in Java/ViewGroups) exists in android :(
Do you have a better idea how to solve this?

JohanDegraeve:
not really, i was just trying to understand the changes

AdrianLxM:
Ah, ok. I was hoping you had better knowledge in the layout-part of Android than I have, @JohanDegraeve.
But StackOverflow concensus also was that you had to generate your own layout.

tzachi-dar:
This looks good to me.
Error messages look better, although background is not white (still much better than before).

I believe that you can merge it.

AdrianLxM:
The background of the error messages in the list will change with the severity level. I did not change that, as it makes the list easier to understand. But I think even the darkest background colour is not too dark and the black text is still readable.

PR #232, opened 2015-12-26
Alerts times6

tzachi-dar:
This is a fix for the alerts timing issues based on the last discussions we had.
I'm running with this for a successful day so far.

Please note that I'll be with very little mail access in the coming days.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
@@ -106,8 +106,6 @@ public class AlertType extends Model {
             return null;
         }
 
-        checkIfMissedReadingAlert(context);
-
         if (bg <= 14) { // Special dexcom codes should not set off low alarms
             return null;
         }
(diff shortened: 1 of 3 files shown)


AdrianLxM:
I forgot to test this, but can't this mess with the reraise interval when a new value comes in?

tzachi-dar:
It should not have any influence on reraised alerts:
The real change in this function is under the code in the function: calcuatleArmTime
In this function, if (activeBgAlert != null) (that is, there is an active alert...) we use the same value as before. Only if this value is negative, we set it to one more minute, so it will probably not create any issues with existing alerts.

AdrianLxM:
Well, what if we reraise every minute, but the system gets triggered a few seconds earlier due to a new value e.g. Than the next alert sound gets stalled by almost a minute because of `if (wakeTime < now + 60000) {...`.

tzachi-dar:
you are probably right, we should probably replace
if (wakeTime < now + 60000) {
with
if (wakeTime < now)

will that solve the issue?

AdrianLxM:
> quoting @tzachi-dar:
> `if (wakeTime < now) ...`

Yes, that's what I was thinking of and this should solve the issue.

It might not be much of an issue at all... only if you don't acknowledge the alarm within 5 minutes or if you just work with large reraise intervals and expect it to be exact. (E.g. reraise value of 30 minutes and it will actually reraise at 31 minutes).
It is more a thing that it could be noticeable to the users and make them loose trust in the alarm system.

PR #235, opened 2016-01-05
Use a new fabric.io key

mgranberry:
Use a new key for crashlytics so we can view crash reports in @StephenBlackWasAlreadyTaken's absence


mgranberry:
Unneeded now that we are part of the organization

PR #237, opened 2016-01-06
Don't crash if a connection is lost/restarted while writing a command

mgranberry:


mgranberry:
This is contained in https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/239

PR #236, opened 2016-01-06
avoid a crash when preparing graph and using comma as a decimal separator

mgranberry:
This should fix this:

```
Fatal Exception: java.lang.RuntimeException: Unable to start receiver com.eveningoutpost.dexdrip.xDripWidget: android.database.sqlite.SQLiteException: near ",": syntax error (code 1): , while compiling: SELECT * FROM Calibration WHERE timestamp >= ,4 ORDER BY timestamp desc LIMIT 288
       at android.app.ActivityThread.handleReceiver(ActivityThread.java:2649)
       at android.app.ActivityThread.access$1800(ActivityThread.java:154)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1398)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:135)
       at android.app.ActivityThread.main(ActivityThread.java:5294)
       at java.lang.reflect.Method.invoke(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:372)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:699)
Caused by android.database.sqlite.SQLiteException: near ",": syntax error (code 1): , while compiling: SELECT * FROM Calibration WHERE timestamp >= ,4 ORDER BY timestamp desc LIMIT 288
       at android.database.sqlite.SQLiteConnection.nativePrepareStatement(SQLiteConnection.java)
       at android.database.sqlite.SQLiteConnection.acquirePreparedStatement(SQLiteConnection.java:891)
       at android.database.sqlite.SQLiteConnection.prepare(SQLiteConnection.java:502)
       at android.database.sqlite.SQLiteSession.prepare(SQLiteSession.java:588)
       at android.database.sqlite.SQLiteProgram.<init>(SQLiteProgram.java:58)
       at android.database.sqlite.SQLiteQuery.<init>(SQLiteQuery.java:37)
       at android.database.sqlite.SQLiteDirectCursorDriver.query(SQLiteDirectCursorDriver.java:44)
       at android.database.sqlite.SQLiteDatabase.rawQueryWithFactory(SQLiteDatabase.java:1316)
       at android.database.sqlite.SQLiteDatabase.rawQuery(SQLiteDatabase.java:1255)
       at com.activeandroid.util.SQLiteUtils.rawQuery(Unknown Source)
[...]

PR #238, opened 2016-01-06
make some timestaps integer

AdrianLxM:
Should solve the same issue as https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/236


mgranberry:
looks like we arrived at the same solution.  LGTM.

PR #239, opened 2016-01-06
fix several gatt-closure crashes for dex share

mgranberry:
There probably needs to be some restructuring here.  All of the null checks make it clear that there's a design problem.


AdrianLxM:
I cannot test this as I have no share. But at the moment it seems reasonable to check for `null`.

Does it need some failover handling when the gatt is null? It looks like this will be handled and hopefully resolved by the retry/failover timers and `onStartCommand` calling `attemptConnection` - but I never took a closer look to how this class works so I'm not sure.

mgranberry:
Yes.  It will be retried by the timer or when the device auto-reconnects if
it is detected before the timer fires, which is what usually happens.

PR #240, opened 2016-01-06
don't crash and burn when making toasts for CSV/database exports

mgranberry:
A bigger problem is that none of the Toasts around errors will actually work (they will crash the app instead).  Should probably be fixed in another PR.


AdrianLxM:
Let me guess, I forgot to run them on the UI-Thread? :facepalm:

mgranberry:
yep.  The road to hell is paved with good intentions.  It looks like posting a toast-creating runnable to the main looper is an easy out for these.

AdrianLxM:
Tested making toast (by causing an error on purpose) and it works great, thanks.

AdrianLxM:
The other changes also look very reasonable to me. I think we can merge this.

PR #241, opened 2016-01-07
don't crash when assigning new share receiver

mgranberry:


AdrianLxM:
Without understanding the protocol: Now it looks like if the assignment is `null` it won't crash anymore but it will not update the monitor assignment and just follow through with whatever the `sessionID` is needed for (creating contacts, deleting contacts, creating invitations, ...).

Is it intended to do so? (Does the assignment not need to be updated in most cases?)
Shouldn't there be an error log that the user can see in the settings?

Sorry for my dumb questions. I've never looked at the share code before and don't even have a system to test it.

mgranberry:
I think you're right.  The protocol is a little bizarre and clearly designed by someone who had never used a web service before.  

The assignment only needs to be updated if it is the first time a receiver is used.  

If the session id isn't available it should probably all fail, but the assignment check sometimes fails without any reason.  I don't have captures of that case, but I tried to replicate the logic that was there before.  This agrees with the flow.

PR #242, opened 2016-01-07
Wear integration

AdrianLxM:
Ported from NightWatch #dc22e2c1d6111784bf647ebb5a0626b64c2f2813
https://github.com/StephenBlackWasAlreadyTaken/NightWatch/commit/dc22e2c1d6111784bf647ebb5a0626b64c2f2813


app/build.gradle
```diff
@@ -102,7 +103,7 @@ dependencies {
     compile 'com.getpebble:pebblekit:3.0.0'
     compile 'io.reactivex:rxjava:1.0.0'
     compile 'com.github.lecho:hellocharts-android:v1.5.5'
-
+    compile 'com.google.android.gms:play-services-wearable:7.5.0'
```
mgranberry, line 106: That's a pretty old version of play services.  Does it still not work on newer versions?

wear/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+}
+
+dependencies {
+    compile fileTree(dir: 'libs', include: ['*.jar'])
+    compile 'com.ustwo.android:clockwise-wearable:1.0.2'
+    compile 'com.google.android.support:wearable:1.1.0'
+    compile 'com.google.android.gms:play-services-wearable:7.3.0'
```
mgranberry, line 27: 7.3 vs 7.5?

AdrianLxM, line 27: This is how they were in NightScout and in this combination and they work. (I tried newer versions, but they didn't on my phone a couple of weeks ago. The cheap chinese 4.4 watches and miniphones can't update to a newer version.)

The only thing I changed was to restrict the play services to "wearable" to not have problems with the DEX method limit.

AdrianLxM, line 27: I'll add it to the list of things I've noticed while porting it - and publish it as a TODO-List in a separate issue.

AdrianLxM:
Here is a list of things I've noticed but didn't want to change right away as they are "stable" in NightWatch and better to do in smaller pull-requests: https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/issues/243

wear/build.gradle
mgranberry, line 27: I was silently worrying about the method limit.  It's a little troubling how easy it is to hit with a few libraries and Google Play Services.  I don't want to think about how much fun it would be to get ProGuard working, but it might be necessary sooner rather than later.  We could probably get away for a while by copying/porting some code out of the android-uploader-core library.

I think the newer version of play services had problems with "legitimate" Android Wear watches.  I was hoping someone had ideas on that.

ghost:
I have successfully tested the wear integration. Works fine with me. 
Under settings "Broadcast locally" can be turned off ("Android Wear Integration" is checked).
All Watch Faces were running without disruption.

Cheers,
Matthias

AdrianLxM:
Thanks @ceben80 for testing it on a Sony Smartwatch 3
Thanks @LadyViktoria for testing it on a Moto 360

It works fine on my Lg G-Watches (tested on the latest update and the version before).

PR #245, opened 2016-01-09
set a vibrate pattern that will never vibrate

AdrianLxM:
_After an OS-Update, Notifications only show on my watch when they either vibrate or a sound is set. In most cases we don't set no sound as we play it separately. Setting the priority to maximum for the notification won't help either._

This change will guarantee that a vibrate pattern is set.
In the case the user doesn't want vibration, the waiting period until first vibration is set to one minute longer that the reraise period. In practice this means no vibration at all - but notifications still come through.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
@@ -267,7 +267,7 @@ public class AlertPlayer {
         } else {
             // In order to still show on all android wear watches, either a sound or a vibrate pattern
             // seems to be needed. This pattern basically does not vibrate:
-            builder.setVibrate(new long[]{alert.getNextAlertTime(ctx) + 60000, 1});
+            builder.setVibrate(new long[]{1, 0});
         }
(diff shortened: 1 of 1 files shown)


mgranberry:
Does this keep the watch awake or something similar?  I'm curious whether battery life is affected.

AdrianLxM:
On a small test on my watch I could not see a faster battery decay than what we had before when it was a standard notification (and was shown).
I'll do some more testing on that and let my spare watch sit with such an alarm over the night for 8 hours or so.

tzachi-dar:
Seems like a nice trick.

AdrianLxM:
Ha, it is much easier without any battery concern: It works, if you wait for a non-zero amount of time (1millisecond is enough) before vibrating and then vibrate for 0 milliseconds.

AdrianLxM:
I've updated the PR to the minimal pattern that does not vibrate.

In the documentation for `public Notification.Builder setVibrate (long[] pattern)` I found: "A notification that vibrates is more likely to be presented as a heads-up notification."
As a developer I really dislike these uncertainties like "more likely".

tzachi-dar:
Probably a nice way to say we have a bug in some version, here is a
workaround.
בתאריך 9 בינו 2016 19:31,‏ "AdrianLxM" notifications@github.com כתב:

PR #244, opened 2016-01-08
Upload sensor data for xdrip viewer mode.

tzachi-dar:
Signed-off-by: Tzachi Dar tzachi.dar@gmail.com

Send data for sensor when in Xdripviewer mode.
Data is needed for better calibration graph.

PR #246, opened 2016-01-09
Upload sensor data for xdrip viewer mode.

AdrianLxM:
Same as https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/244
but using LF instead of CRLF.


tzachi-dar:
Thanks for the pr.
בתאריך 9 בינו 2016 21:07,‏ "AdrianLxM" notifications@github.com כתב:

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -366,7 +390,9 @@ private boolean doMongoUpload(SharedPreferences prefs, List<BgReading> glucoseDa
 
                         for (Calibration calRecord : calRecords) {
                             //do not upload undefined slopes
-                            if(calRecord.slope == 0d) break;
+                            if(calRecord.slope == 0d) {
```
AdrianLxM: good catch!

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -366,7 +390,9 @@ private boolean doMongoUpload(SharedPreferences prefs, List<BgReading> glucoseDa
...
+                                continue;
```
AdrianLxM, line 394: good catch!

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -388,6 +414,20 @@ private boolean doMongoUpload(SharedPreferences prefs, List<BgReading> glucoseDa
...
                         }
+                        
+                        Log.i(TAG, "The number of sensor records being sent to MongoDB is " + meterRecords.size());
+                        for (Sensor sensor : sensors) {
+                            // make db object
+                            BasicDBObject testData = new BasicDBObject();
+                            testData.put("type", "sensor");
+                            testData.put("xDrip_started_at", sensor.started_at);
+                            testData.put("xDrip_stopped_at", sensor.stopped_at);
+                            testData.put("xDrip_latest_battery_level", sensor.latest_battery_level);
```
AdrianLxM, line 425: This will get updated when a new transmitter package comes in with a lower battery level.
I guess it would suffice adding the whole Sensor to the SensorSendQueue again since we switched to upserts.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/SensorSendQueue.java
```diff
@@ -22,23 +23,18 @@
...
+    
+    public static List<SensorSendQueue> mongoQueue(boolean xDripViewerMode) {
+        List<SensorSendQueue> values = new Select()
                 .from(SensorSendQueue.class)
-                .where("success = ?", false)
                 .orderBy("_ID desc")
+                .limit(xDripViewerMode ? 100 : 0)
                 .execute();
+        if (xDripViewerMode) {
+             java.util.Collections.reverse(values);
```
AdrianLxM, line 34: If we just upload in xDripViewerMode, the order could be set in the query.
But from my standpoind we could very well upload the data in any case. A new Sensor is set every couple of days; maybe when the battery decays we would need some more updates but is still not very much and would not overload any connection.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
AdrianLxM, line 425: The battery level gets updated in `DexCollectionService` line 510 and `WixelReader` line 509.

AdrianLxM:
@tzachi-dar, just while you're at it: `Sensor.java` is in the main package. Could you please refactor/move it to package `Models` if it is not too much work.

tzachi-dar:
Ok, did all things that you have mentioned.
Also added the alerts time fix so I can check them together.
Opened a new PR that is based on the new master.

PR #247, opened 2016-01-09
Sensorupload2

tzachi-dar:
Sensor update on new master, with remarks from previous branch.


AdrianLxM:
Thanks for this and also for the changes. It looks good to me :)
(I can't await my sensor change on Monday morning now ;) )

AdrianLxM:
I tested the alarm timer modification - works fine;
I did the whole new sensor process: It stopped and started as it should; battery_level seems to be set correctly with the new function. :+1:

PR #252, opened 2016-01-15
Proguard

AdrianLxM:
This is the very naive approach adding nowarns for anything.
The retrofit rules already were in place but I couldn't test them.

So far I successfully tested it for:
- Wixel input methods
- Alarms
- Widgets (home and lockscreen)
- Android wear / smartwatch
- Direct Mongo upload
- Speak Readings
- Most settings
- Statistics

It would be good if this could be tested for the Share-Receiver, Pebble and Share and Nightscout upload.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -88,6 +88,11 @@ android {
         multiDexEnabled true
     }
     buildTypes {
+        release {
+            minifyEnabled true
+            proguardFiles getDefaultProguardFile('proguard-android.txt'),
+                    'proguard-rules.pro'
+        }
     }
 }
 
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -20,3 +20,11 @@
 -keep class retrofit.** { *; }
 -keepattributes Signature
 -keepattributes Exceptions
+-dontwarn com.**
+-dontwarn net.**
+-dontwarn okio.**
+-dontwarn org.**
+-dontwarn rx.**
+
+
+-dontobfuscate


tzachi-dar:
2 questions, 
1) Did we always have proguard and we are only extending this now?
2) I thought that we either should use multiDex or Proguard. Should we use both?

I

AdrianLxM:
1) We never used proguard. @mgranberry had just been wisely suspecting we could need it and added rules for retrofit. (That uses reflection)

2) I think we'd choose one of them. Proguard has the advantage of quicker startups, smaller .apks and less memory usage. We have a working solution with multidex so there is no rush. I'd really like to have everything tested with proguard before choosing one or the other.

Proguard seemd to have an advantage for XDripViewer to be able to address older Android versions. (xdripviewer -> no BLE -> min target can be lower). But I've just read, that the multidex compatibility library now also supports older devices. I didn't have time to read the experience reports though.

I'd say: no rush, but it would be great to test the other parts in order to have the option.

mgranberry:
Those are pretty broad exclusions.  I know the android-uploader-core JAR will cause a lot of unresolvable methods to exist, but proguard can also point to legitimate bugs.  I usually try to exclude packages instead.

AdrianLxM:
@mgranberry, if I understand you right, we would need those exclusions anyways for it to work now.
I totally understand that it would be better to exclude the single packages - if we add something new that has legitimate bug warnings the risk is quite high that they get masked by the broad exclusions.

Before I do the work and exchange the general exclusions for more specific ones I'd like to know if we would switch to proguard. So it still would be great if you could test it with Nightscout and Share (upload and Receiver).

tzachi-dar:
so, if I understand correctly, I should try with this and report what is going on (I'll test that with 4.2) builds.

tzachi-dar:
By the way, using multydex causes the build to take a longer time (or was this the watch-face?)

mgranberry:
Multidex, proguard, and the watchface will all increase build times.  Moving to a newer version of Gradle and Android build tools might help some.

AdrianLxM:
> so, if I understand correctly, I should try with this and report what is going on (I'll test that with 4.2) builds.

That would be great!

tzachi-dar:
So, tough luck with the 4.2 testing...

I still don't know if it is proguard related or multidex related, but when I open the navigation bar, the screen will flicker for a long time.

I have seen this also happening with gmail, but never before it has happened with xDrip.

I'll check tonight what is the CL that has caused it and will see how to proceed.

Thanks
Tzachi

tzachi-dar:
Ok,

So after leaving the table like this for the entire day, the problem seems to be solved.
I have removed the build and installed it, also rebooted a few times, and this does not seem like a real problem.

In a day or two, I'll probably install it on another tablet and try to create a movie of that.

So, far, seems that we need to keep following, but the problem seems to be gone.

Tzachi

tzachi-dar:
Tested this today on another tablet with 4.2.
Works well.

Also continuing to work well on the previous tablet.

I'll continue the follow up.

AdrianLxM:
I think we won't use proguard after all and if so we will need a new ruleset. I'll close this PR.

PR #253, opened 2016-01-15
wakelock: try-finally

AdrianLxM:


app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
```diff
@@ -97,14 +97,17 @@ public Notifications() {
...
+            wl.acquire();
+            Log.d("Notifications", "Running Notifications Intent Service");
+            Context context =getApplicationContext();
+            ReadPerfs(context);
+            notificationSetter(context);
+            ArmTimer(context);
+            context.startService(new Intent(context, MissedReadingService.class));
+        } finally {
+            if (wl.isHeld()) wl.release();
+        }
```
tzachi-dar, line 110: Usually, we do the acquire outside the try block.
Seems to me that this should also work, but if we want to be more consistent, than we should change.

AdrianLxM:
Ok, I moved it as I have no preference for one over the other. Not yet tested it.
As long as @mgranberry 's linter doesn't complain ;)

mgranberry:
Looks good to me.  Thanks. :)

PR #255, opened 2016-01-15
Also add date for sensor.

tzachi-dar:

Needed for the NS api to allow reading this.


tzachi-dar:
Also added send the calibration to the sendqueue also when it is being marked as not important.

PR #258, opened 2016-01-22
layout_below instead of many different margins

AdrianLxM:
This works on my phone to automatically adjust the chart size and position when adding or removing the extra status line. Also when the "signal missed" line gets added.


tzachi-dar:
Looks great... merging it.

AdrianLxM:
sorry, restoring branch... will delete it in the next couple of days.

PR #262, opened 2016-01-25
widget range lines

AdrianLxM:
Introduces a setting that can enable/disable the range lines on the widget.


tzachi-dar:
Looks good to me.

PR #263, opened 2016-01-25
BG History

AdrianLxM:
This will add an Activity ("BG History" in the navdrawer) that shows the bg values and calibrations of a selected day (zoomable, similar to home).


AdrianLxM:
Like @tzachi-dar wished, I've added the possibility to define a range of days.

I tested it on longer and shorter days (daylight saving time adjustments) and it works fine. It will just have either one hour mark doubled (summer -> standard time) or one hour mark left out (standard -> summer time). The chart plots as it should.

tzachi-dar:
Thanks.
בתאריך 26 בינו 2016 17:16,‏ "AdrianLxM" notifications@github.com כתב:

tzachi-dar:
Just looked at the graphs created. Looks really nice.

I haven't still looked at the code yet.

It seems that I was not understood well. The graph that I actually need is a few days on one day graph.  This way they can see how bg behaves by the hours.

Here is an example from dexcom studio:
image

In any case, what was done here is great...

AdrianLxM:
A couple of colleagues have tested it (thanks Barbara, RTox, Ceben, ...) and it seems to work fine on 4.4, 5.1 and 6.0.1 Android.
Before I merge it I'd still like to have a good to go from one of the core developers, so: Can I merge this?

tzachi-dar:
I have looked at it as a user, but I still want to look at the code, so
please wait for tonight.

Thanks
Tzachi

tzachi-dar:
OK, looked at the code. Not an expert on UI, but this seems good to me.

PR #268, opened 2016-02-02
Fix for cascading alerts.

tzachi-dar:

This is a fix for adrians 2 alerts problem. Not tested yet.


AdrianLxM:
Works great! I think you can merge this. Thanks!

PR #269, opened 2016-02-04
widget extra status line

AdrianLxM:
Add a setting that will also show the extra status line on the widget.
screenshot_2016-02-04-12-30-50

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -279,7 +279,7 @@ public class Home extends ActivityWithMenu {
             notificationText.append("\nHIGH ALERTS CURRENTLY DISABLED");
         }
         if(prefs.getBoolean("extra_status_line", false)) {
-            extraStatusLineText.setText(extraStatusLine());
+            extraStatusLineText.setText(extraStatusLine(prefs));
             extraStatusLineText.setVisibility(View.VISIBLE);
         } else {
             extraStatusLineText.setText("");
@@ -425,7 +425,7 @@ public class Home extends ActivityWithMenu {
     }
 
     @NonNull
-    private String extraStatusLine() {
+    public static String extraStatusLine(SharedPreferences prefs) {
         StringBuilder extraline = new StringBuilder();
         Calibration lastCalibration = Calibration.last();
         if (prefs.getBoolean("status_line_calibration_long", true) && lastCalibration != null){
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -425,7 +425,7 @@ private void displayCurrentInfo() {
     }
 
     @NonNull
-    private String extraStatusLine() {
+    public static String extraStatusLine(SharedPreferences prefs) {
```
tzachi-dar, line 427: Please rename this variable (perfs).
Otherwise, we will have two variables one class member and one local with the same name.

By the way, it might be good to rename the class member to perfs_ and live it as is.

AdrianLxM, line 427: One is an instance/member variable the other one a parameter of a static function. No clash there.
But still: Android style would be calling the member `mPreferences`. And we could really use a bit more readable code. I'll change that tomorrow.

AdrianLxM, line 427: I renamed the member.

tzachi-dar:
Strange thing, I don't see this data. My widget is also smaller than the one that you are using.
By the way, playing with the UI it seems that this will only work on the widget if it is also on the main screen. Is this by design?

AdrianLxM:
If you have the widget smaller it will be cut off (like the minutes ago is cut off). If you make the widget bigger, it should be visible.

And yes, that's how It si programmed: the widget will just show the status line, if the general status line switch is switched on. It can easily be changed but with the drawback that there won't be a general status line switch anymore that will grey out the elements in the settings if switched off. I have no preference.

tzachi-dar:
OK, played with it, looks nice.
Only one question left: do we want the status line on the middle of screen, or maybe try to put it on the bottom of widget or top of it?
In any case this looks good to me.

AdrianLxM:
I like it where it is, but tastes are different. And "below the minutes ago line" seemed practical to me.
I guess where it is now it is less likely to get cut off on smaller widgets - and we cannot predict how big the user will have the widget - so we have to do it dynamically I guess (and still make sure it won't get drawn on top of the other text elements?). And the size can be different for different Home-screens and Lock-Screens on the same device.
If you find a quick solution (or the time for a not so quick one) feel free to change the layout.

tzachi-dar:
Let's commit it as is, and once I'll have time I'll see how I can change
that.

Thanks
Tzachi

PR #271, opened 2016-02-04
Fix false low notifications

AdrianLxM:
People were complaining on Facebook (and here a few month back) that XDrip alerted for an empty transmitter even though it was quite new.

Some transmitter seem to send impossibly low battery values from time to time and therefore cause the alert. With this PR we will just accept battery levels above 120. (This is much lower than I have ever seen a transmitter being able to transmit anything).


tzachi-dar:
Seems reasonable.

PR #272, opened 2016-02-08
wear: match default values in settings and code

AdrianLxM:
The watchfaces should have a black background by default. The default value was just set in the settings but not in the code. Without changing it in the settings at least once the settings did not represent the actual behaviour.


tzachi-dar:
LGTM.

PR #273, opened 2016-02-11
Restrict data fetched for ongoing Notification

AdrianLxM:
This restricts the data fetched for the ongoing notification to the data that is needed.
It will not fetch the whole day but just the 3 hours shown in the notification.

This also fixes the bug that values hours before the values shown in the chart affect how it is stretched on the y axis. Thanks @fezulin for reporting this.


AdrianLxM:
Sorry, there is more to do. Also on the updates.

AdrianLxM:
Now it also works for updates :)

fezulin:
Perfectly done! Thank you!
screenshot_before_fix
screenshot_after_fix

PR #279, opened 2016-02-16
Settings ex import

AdrianLxM:
This is @jamorham 's im-/export functionallity (https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/266) with some small changes:
- export settings to a subdirectory of the one we use for csv and database exports
- no recursive deletion of the export directory
- added Nav-Drawer to the Import/Export-Activity instead of floating close button
- adopted Import/Export-Activity to small screens
- grouped all export/import items in the menu

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -220,6 +220,10 @@
         <activity
             android:name=".FollowerManagementActivity"
             android:label="Manage Followers" />
+        <activity
+            android:name=".utils.SdcardImportExport"
+            android:noHistory="true"
+            android:label="Settings on external storage" />
 
         <service
             android:name=".Services.SnoozeOnNotificationDismissService"
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -44,6 +44,7 @@ import com.eveningoutpost.dexdrip.UtilityModels.Intents;
 import com.eveningoutpost.dexdrip.stats.StatsResult;
 import com.eveningoutpost.dexdrip.utils.ActivityWithMenu;
 import com.eveningoutpost.dexdrip.utils.DatabaseUtil;
+import com.eveningoutpost.dexdrip.utils.SdcardImportExport;
 import com.eveningoutpost.dexdrip.wearintegration.WatchUpdaterService;
 import com.nispok.snackbar.Snackbar;
 import com.nispok.snackbar.SnackbarManager;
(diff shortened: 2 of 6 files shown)


jamorham:
This looks good to me. `ScrollView` is a good addition, menu layout looks sensible to integrate with existing import/export features and consistent directory path with `getExternalDir()`.

tzachi-dar:
Nice job jamorham and AdrianLxM.

2 things that I think need to be changed:

1) "Settings to external storage". I think it needs to show that is also "Settings from external storage". Not sure if this can be done given the limited space.

2) This activity works on the file /sdcard/xdrip/settingsExport.
Since people might use this to pass xDrip from phone to phone, I believe that the text should mention this file location.
For example: "The settings will be stored/restored from the directory "/sdcard/xdrip/settingsExport.  Please make sure it is there before restoring".

Also when setting, the text used is sdcard'/xdrip/SettingsExport'. This should be '/sdcard/xdrip/SettingsExport'

Thanks
Tzachi

AdrianLxM:
@tzachi-dar 

ad 1) you click "Im/Export..." and then "Settings to external storage" we could make that:
 "Im/Export..." and then "Im/Export Settings"

ad 2 and 3) It actually works on the same external storage directory where the databases (and logcats?) are exported to. So "sdcard" is not necessarily part of the path.
On modern phones (that we all use) there always is an external storage - if not an sdcard it will be the built in external storage. Just some Android distros allow to restrict the access - but then the user knows what he is doing.

But actually I don't care much what the text says: Just tell me what you would like and I'll put it ;)

tzachi-dar:
As for (1) go with it :-)

As for 2, please add:
"The settings will be stored/restored from the directory '" + getCustomSDcardpath() +  "' +/sdcard/xdrip/settingsExport. Please make sure it is there before restoring".

As for 3, please replace:
toast("Preferences saved in sdcard '/xdrip/settingsExport' ");
toast("Preferences saved in ' " + getCustomSDcardpath() +" ' " ); // remove the spaces, I have added a space between ' and " because otherwise it is unreadable

AdrianLxM:
I think that something like "/storage/emultated/0/xdrip/settingsExport" will confuse people as most android file managers won't show it like this. Neither does it show like this when connecting to a PC via data cable.
But ok, I'll put it like this - people will ask of Facebook anyways.

tzachi-dar:
I believe that if this is what it will be showing, than don't change what you have...

AdrianLxM:
Ok, I updated the menu names.
Also added the directory path to the "could not open..." toast.

PR #282, opened 2016-02-19
Multilanguage support

viktor-r-r:
Some text has been transferred to "strings.xml"
and has been translated into Russian (values-ru)


tzachi-dar:
Hi, 

It seems that there are many files that are not related to the translation, and should not be really checked in.

Looking at the zip file that you have sent, it seems that you have added 3 files in the directory:
xDrip-multilangsupp.zip\xDrip-master\app\src\main\res\values-ru

Are there more files that should be added?

tzachi-dar:
Hi,

I went over the pr that you have sent and also the zip file that you have sent me.

It seems that this are based on very old xDrip version, and are therefore not usable.

In order to be able to translate xDrip to ru and also many other languages, I advise that you will do the following:

1) update to the latest xDrip master code.
2) Create a new PR that is based on this code.

My advise is that you will start with smaller (and simpler) PRs that we can review and merge fast.

Does this sound reasonable?

AdrianLxM:
Yes, it looks like it is based on a version that is around commit number 220. For comparison the current master has 1149 commits.

I actually like what you did @viktor-r-r, unfortunately it is not mergable. Could you please base this on the current master of xDrip-Experimental. The basics of Git are not hard. The basic steps are
- Fork the repository - on Github
- Clone your fork to your local system - git console `git clone ...`
- Do the changes
- call `git add ...` on the files you've changed
- call `git commit`
- call `git push` to push things back to github
- on github do a pull request

PR #285, opened 2016-02-23
Fix for issue ascending alert switches of the radio/mp3.

tzachi-dar:
The root cause of this bug was that we have been creating the media player on a thread that was dying soon after.

The solution is to call it from the UI thread.

This was heavily tested on 4.4.4. Will test it on 5.1 tomorrow. 

This pr still has some printing and it also uses very low volume, in order not to drive me nuts when testing it.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
@@ -63,7 +63,7 @@ class MediaPlayerCreaterHelper {
         
         try {
             synchronized(lock1_) {
-                while(mplayerCreated_ = false) {
+                while(mplayerCreated_ == false) {
                    
                         lock1_.wait();
                 }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
```diff
@@ -19,8 +22,64 @@
...
+                    }
+                    
+                }
+            }
+        };
+        mainHandler.post(myRunnable);
+        
+        try {
+            synchronized(lock1_) {
+                while(mplayerCreated_ == false) {
```
AdrianLxM, line 66: Can we be sure here, that we are not on the same Thread that the mainLooper is on, what could lead to a race condition where we block the runnable (mediaPlayer-producer) from being executed? (I don't know if one thread can have several loopers or message queues.)

tzachi-dar, line 66: This code will only run if we are not on the ui thread.
See the function isUiThread().

tzachi-dar:
By the way, any ideas how to debug the anr problem? (reported in the bug)?

AdrianLxM:
Please tell me, when the pr is in the final state. I'd like to test it on a live system.

tzachi-dar:
OK, you can start testing.

AdrianLxM:
Works fine on my test system. I'd merge it.

PR #286, opened 2016-02-25
Make sure that the hour and date pickers are combined.

tzachi-dar:

Simple code to make sure that on start sensor the two pickers work together.

This is something that was annoying me from day one, happy to finally fix it.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
@@ -23,7 +23,18 @@ public class StartNewSensor extends ActivityWithMenu {
     private Button button;
     private DatePicker dp;
     private TimePicker tp;
+    
+    private int last_hour;
 
+    void AddDays(int numberOfDays) {
+        Calendar calendar = Calendar.getInstance();
+        calendar.set(dp.getYear(), dp.getMonth(), dp.getDayOfMonth(),
+        tp.getCurrentHour(), tp.getCurrentMinute(), 0);
+        
+        calendar.add(Calendar.DAY_OF_YEAR, numberOfDays);
+        dp.updateDate(calendar.get(Calendar.YEAR), calendar.get(Calendar.MONTH), calendar.get(Calendar.DAY_OF_MONTH));
+    }
+    
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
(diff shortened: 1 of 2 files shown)


AdrianLxM:
This is a behaviour I would really dislike.

First I choose the date, then I choose the time... and then I have to correct the date again as the time picker changed the date. With the additional problem that you can't see the when you have scrolled down to the hour picker on small devices you can't see the date picker.

app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
```diff
@@ -23,7 +23,18 @@
     private Button button;
     private DatePicker dp;
     private TimePicker tp;
-
+    
+    private int last_hour;
+
+    void AddDays(int numberOfDays) {
```
AdrianLxM: this should be `addDays`

AdrianLxM:
A possibility could be to have the time picker above the date picker?

tzachi-dar:
OK, replaced the order, time is now first.

Please let me know if this looks good.

AdrianLxM:
Today I would have fallen for this trap again :/

I sometimes start the sensor just a bit over 24 hours after insertion - sometimes I wait 48 hours until I switch transmitters.
The bad translation and handling of "am" and "pm" and that 16:00 can be before 10:00 (and therefore cause a day jump) certainly has it's influence. Could you make the spinner adopt to 24 hour format, please?
If I'm not mistaken, `tp.setIs24HourView(android.text.format.DateFormat.is24HourFormat(getApplicationContext()));` in `onCreate()` should suffice. (Not tested, not written in an IDE.)

tzachi-dar:
I'll change that to 24 hours format, and if you fill it is still problematic I can add a checkbox saying link the date and time pickers.
Since it will only be in this dialog it will not add anything else to the settings.

tzachi-dar:
OK, made the time viewer use 24 hour format.

AdrianLxM:
I'm fine with always 24h format. But how about the US guys?
`android.text.format.DateFormat.is24HourFormat(getApplicationContext())` should return `true` if the user setting in the phone are 24h format and `false` otherwise.

tzachi-dar:
I thought that you wanted to always have it to 24 hours.
what are the setting on your phone?

tzachi-dar:
OK, fixed it to match the settings on the phone.

tzachi-dar:
Ok, seems that setting the picker to use 24 hours format was causing a crash when rotating the phone when the dialog was opened.

following http://stackoverflow.com/questions/25880638/rotating-android-device-while-viewing-dialog-preference-with-timepicker-or-numbe, the fix was trivial.

Also added a button to make this entire "feature" optional.

Please let me know if there are any more comments.

PR #296, opened 2016-03-16
Fix crash when serice gets stopped and restarted by the system

AdrianLxM:


tzachi-dar:
Seems reasonable, although I'm not familiar with wear code.

PR #297, opened 2016-03-22
Add snooze action to BG alerts and increase priority to MAX

sachams:
- Adding the snooze action makes it much easier to snooze the alarm, especially in the middle of the night
- It also shows up on your Wear watch
- Increasing the priority to PRIORITY_MAX means that you can turn off notifications at night, apart from priority BG alarms
- It also means that the notification will definitely show up - on some devices only PRIORITY_MAX notifications will actually display a HUD notification.


tzachi-dar:
Going directly into snoozing, makes snoozing much simpler, but you don't really know what you are snoozing. (you can not see the graph, or even the bg value itself).
Can you make that configurable so that everyone will have the possibility to use that or not?
(please add one more checkbox on "bg alerts setting").

Thanks
Tzachi

sachams:
good point! Sure, will do.

AdrianLxM:
I'm on holidays so it might take some time ( at least 2 weeks) until I fully can look this through, sorry.

Could you please explain what will change?
At the moment you can simply swipe away the notification on the wear watch and it will snooze the alarm for the default period. Can you now swipe to the left and press "open on phone" to snooze it without waiting the time you can restore the "Notification card" on the watch (or doing the famous swipe right+up+right combination for quick dismissal)?

The `PRIORITY_MAX` did not work for me. I therefore introduced a not noticeable vibration pattern what made the Notification always appear.

sachams:
Hi,

Snooze action change
I found that my phone would make a noise/vibrate but no notification would be displayed. I had to open xDrip and go to the snooze page to snooze it. This change adds a snooze action to the notification so you can just press the button on the notification to do the default snooze. On the Wear watch you can swipe to the left and click on the snooze button to snooze. Swiping the notification on my watch didn’t seem to halt the alarm.

PRIORITY_MAX change
If you want to get BG alerts during the night, but don’t want to get alerts when you get a new SMS, then currently you have to turn off all the things you don’t want to get alerted about, and leave the xDrip alert on. This change makes the priority for BG alerts to MAX. This means you can set “only show priority notifications” on your phone between (e.g.) 22:00 and 07:00 and you get BG alerts but not texts.

To make PRIORITY_MAX appear on al devices you also have to set a vibrate pattern - just setting PRIORITY_MAX by itself doesn’t seem to work :-/ So you might be able to use a simple notification with PRIORITY_MAX and a repeating vibrate pattern.

AdrianLxM:
Thanks a lot @sachams for the PR and the explanation :)

From the explanation it looks like a good addition to me. Swiping the notification halts the alarm but only after the time in which you can restore the notification on the watch has run out.This can be a long time during the night though if you don't want to wake your spouse. I've formed the habit of swiping the alarm, semi-restoring it (just until the restore button) and then swiping the button... this speeds up things.
But a left-swipe+button press is definitely more convenient.

@tzachi-dar
I believe that we don't lose any functionality this way (You can still do it the conventional way - it is "just" an added button, isn't it?). And from the notification text we can also see what alarm we are "quick snoozing".
I will check (especially the wear part) after the Easter Holidays.

tzachi-dar:
@Adrian, if I understand that correctly, on a table that is always open with xDrip this means that on alert, the screen will change from displaying the graph to displaying the snooze activity.
Personally, it is more important to me to see the graph in the case of alert, so I have asked to make that configurable.

sachams:
cool. I might have a look at the alert text too, to make sure we can convey the best info in the alert, so that when you snooze you know what you are snoozing.

And for me too the driver was that if it happens too much your partner is going to get really annoyed!

tzachi-dar:
The way I understand it, when alert is played, the dialog will change to the snooze.
I'll try that tonight to see how that actually works.

By the way, as for the partner issue, well, I sometimes pretend to sleep wait for her to wake up...

tzachi-dar:
Didn't have the time to test that today, will try to do that tomorrow.

tzachi-dar:
OK, looked at it today more deeply.

Basically there are 3 changes here:

1) raising alert priority.
2) Adding a button to the bg alerts.
3) Adding a button to the snooze alert.

# My response:

0) There is also the calibration alert that needs to have the same treatment as the other alerts.
1) raising alert priority, I want to play with it some more to understand it better.
2) Adding a button on bg alert. Well, this looks good. I don't have an objection to adding it.
3) Other alerts:
  a - on a phone, it is only possible to snooze the alert with the default snooze. So, on a phone I don't think that this button is needed. If it is needed for a watch, then let's add it.
  b - In order to snooze these alerts, there is no need to open the snooze activity. Actually, this alerts are being created already snoozed, so you just need to silence them. running the snooze activity in this case is not needed and you only need to silence them. So in the case of other alerts, you need to open the home activity directly.

Does this make sense?

Thanks
Tzachi

PR #299, opened 2016-03-22
Missed reading alert

tzachi-dar:
This PR makes the missed reading alert time dependent.
That is one can set it to work only on day or only on night.


tzachi-dar:
Just added one more important fix to this branch.
It seems that we have this bug in more places, after this is merged, I'll fix the other places.

The problem happens when one enters "" (that is nothing) instead of a number and then the function Integer.parseInt crashes. Not really related to this commit.

tzachi-dar:
Merged this against latest master.

AdrianLxM:
can we close this?

PR #300, opened 2016-03-23
Add timeout on mongo send/receive operations.

tzachi-dar:

This is a fix for the problem of mongo upload operations that got stuck.
It seems that we have not configured timeout on the operations and as a result, this might got stuck for ever.
Verified the fix is working with much smaller timeouts.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/MongoWrapper.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/MongoWrapper.java
@@ -36,7 +36,7 @@ public class MongoWrapper {
 	// Unfortunately, this also throws other exceptions that are not documetned...
     public DBCollection openMongoDb() throws UnknownHostException {
 
-    	MongoClientURI dbUri = new MongoClientURI(dbUriStr_); //?? thros
+    	MongoClientURI dbUri = new MongoClientURI(dbUriStr_+"?socketTimeoutMS=180000");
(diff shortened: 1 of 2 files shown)


jstevensog:
I think merge away.
Nice catch Tzachi-dar.
Cheers

---- tzachi-dar wrote ----

AdrianLxM:
Nice catch.
Before merging, could you please change the line separators.

tzachi-dar:
Sure, I''' change the line separators.
I want to run with it for a couple of more days before merging it, so I will have more confidence in this change.

aaroecker:
Could a time out be created for the collector?  There's a bug that's been talked about at length concerning loss of Bluetooth signal.  When the restart collector button is pushed the Bluetooth icon reappears on the Dex.  If no data is collected by the collector after six minutes a trigger could automatically restart the collector hopefully solving the issue.  Just a thought.

jstevensog:
I have tried this in the past.  But not been successful.  Will try to revisit it soon.
Cheers

---- aaroecker wrote ----

aaroecker:
Awesome, thank you!

tzachi-dar:
Merged into master (not through github).

AdrianLxM:
What did you actually change? Due to the line endings it looks like you switched the whole file.

tzachi-dar:
Oops, only changed 2 lines, fixing that soon.

tzachi-dar:
OK, fixed that, look at maste again.

PR #314, opened 2016-03-31
Make sure that alerts will also be active for very low packets.

tzachi-dar:

This fixes the issue of no alerts when packets get very low.


tzachi-dar:
Merged in 327.

tzachi-dar:
Actually was merged in 329.

PR #316, opened 2016-03-31
Make the Handling of TransmitterData more consistent with rest of the…

tzachi-dar:
… code.

This is basicly code refactoring with no side effects.

One thing that has changed, is that keepAlive() will now be called even if the packet is duplicate.
I believe that this is better this way. If this is not the case then the function processNewTransmitterData should return bool telling if this is a duplicate packet.

Not tested for obvious reasons....

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/G5CollectionService.java
@@ -109,7 +109,6 @@ public class G5CollectionService extends Service {
     private BluetoothDevice device;
     private long startTimeInterval = -1;
     private int lastBattery = 216;
-    private long lastRead = new Date().getTime() - (5 * 60 *1000);
     private Boolean isBondedOrBonding = false;
 
     private static final ScheduledExecutorService worker =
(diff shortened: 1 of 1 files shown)


AdrianLxM:
Deduping is now done at another place or is there no deduping at all?

So basically the change is: Waking up to keep the device alive for 20 seconds now happens about 5 minutes after the **last** package of a number of duplicates arrives instead of about 5 minutes after the **first** package?

I think @urmc-costik (or @hackingtype1 ?) should decide if that makes sense.

app/src/main/java/com/eveningoutpost/dexdrip/Services/G5CollectionService.java
```diff
@@ -563,8 +553,8 @@ private void processNewTransmitterData(TransmitterData transmitterData, long tim
         Sensor.updateBatteryLevel(sensor, transmitterData.sensor_battery_level);
         android.util.Log.i("timestamp create", Long.toString(transmitterData.timestamp));
 
-        BgReading.create(transmitterData.raw_data, transmitterData.filtered_data, this, transmitterData.timestamp);
-        transmitterData.save();
+        BgReading.create(transmitterData.raw_data, filtered_data, this, transmitterData.timestamp);
+
 
     }
```
hackingtype1, line 560: The last read, time since logic should probably stay put. When the G5 wakes up, the app may acquire raw data multiple times before it goes back to sleep. I put those checks in place because BgReading create was inserting multiple duplicates.

tzachi-dar:
Deduping now happens inside TransmitterData.create(). It also checks more things there.
Also, even on the first time that we wake up, we know what was the last packet.

The real question is weather, keepAlive() logic depends on the fact that this is a duplicate packet or not.

app/src/main/java/com/eveningoutpost/dexdrip/Services/G5CollectionService.java
jr-costik, line 560: Ah, just saw your note Adrian - is deduping occurring elsewhere? I haven't seen it happening, which is why I implemented it that way. I would hold off on these changes for now, revisit once we get more stable process and startup with G5 users.

---

John Costik :: Lead Software Developer
Center for Clinical Innovation :: University of Rochester
(585) 350-5391tel:(585)%20350-5391

On Thu, Mar 31, 2016 at 3:21 PM -0700, "John Costik" <notifications@github.com<mailto:notifications@github.com>> wrote:

In app/src/main/java/com/eveningoutpost/dexdrip/Services/G5CollectionService.javahttps://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_StephenBlackWasAlreadyTaken_xDrip-2DExperimental_pull_316-23discussion-5Fr58137513&d=BQMCaQ&c=4sF48jRmVAe_CH-k9mXYXEGfSnM3bY53YSKuLUQRxhA&r=-S4tVkPALJfUIaQ0q618B_o0Ok9zeQjeO8H8WZ2UpSA&m=bEC20KGYzyKaAP4uY6483hLfBtsaB8g1JYSEg_aQBTg&s=MEaTKPdDEs7BLXbNx5fDmQAlYI8_5X0drXWWS-skICQ&e=:

> @@ -563,8 +553,8 @@ private void processNewTransmitterData(TransmitterData transmitterData, long tim
>          Sensor.updateBatteryLevel(sensor, transmitterData.sensor_battery_level);
>          android.util.Log.i("timestamp create", Long.toString(transmitterData.timestamp));
> -        BgReading.create(transmitterData.raw_data, transmitterData.filtered_data, this, transmitterData.timestamp);
> -        transmitterData.save();
> - ```
>      BgReading.create(transmitterData.raw_data, filtered_data, this, transmitterData.timestamp);
>   ```
>   
>   +
>   
>    }

The last read, time since logic should probably stay put. When the G5 wakes up, the app may acquire raw data multiple times before it goes back to sleep. I put those checks in place because BgReading create was inserting multiple duplicates.

hackingtype1:
Let me test it out tonight; if deduping is working - I'll verify the keep alive is still triggered appropriately

tzachi-dar:
OK, let us know how it is going.

AdrianLxM:
I'm closing this as it is already in https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/317

PR #317, opened 2016-03-31
dupe & transmitter data, hold on alpha update

hackingtype1:
This should superscede https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/316#issuecomment-204154549

The null check on Transmitter data was incorrect, and I've adjusted the keep alive timer to prevent missing the G5 wake up. All other changes are the same.

a few other changes for proper bond/unbond handling


tzachi-dar:
Looks good to me, merging this.

PR #318, opened 2016-04-01
pause before reading, stop scan w/ sensor stop

hackingtype1:
Inducing a brief delay to test improved receiver/xDrip cooperation & reliability in getting data. Current theory is that they both begin requesting data when the G5 begins advertising.. colliding & preventing the receiver from getting data.

Also, when the sensor session is ended, the G5 service will stop scanning. This will prevent unwanted re-auth/bonding when a user decides they no longer wish to use xDrip.

also fixed issue presented in crash data (stopScan request when not scanning)

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/G5CollectionService.java
@@ -110,6 +110,7 @@ public class G5CollectionService extends Service {
     private long startTimeInterval = -1;
     private int lastBattery = 216;
     private Boolean isBondedOrBonding = false;
+    private Boolean isFirstTry = true;
 
     private static final ScheduledExecutorService worker =
             Executors.newSingleThreadScheduledExecutor();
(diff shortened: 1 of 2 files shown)


tzachi-dar:
One note about stopping the sensor:

There are two reasons that I think that this should not be in the final version:
1) All other collection methods keep collecting even when sensor is stopped.
2) Stopping/starting a sensor also has influence on calibrations.

So, I think that for final version we will have to give users a special button to stop the scanning if they need to.

AdrianLxM:
@tzachi-dar 
Apart from storing it into the db, we basically do nothing with transmitter data while no sensor session is active. For the G4 it is ok to collect the data - it broadcasts it anyways so it won't affect the transmitter battery.
For the G5 I think it's a good Idea to not "contact" it when not really needed to not waste the transmitter battery.

A switch to manually disable the collection of G5 data without having to reenter the transmitter ID to reenable it ("forget device" atm deletes the transmitterID) would be good for the final version (but I see this as low priority as it can be added by any of us).

PR #321, opened 2016-04-03
alpha 6

hackingtype1:
added missing else statement for sensor start


hackingtype1:
Seeing some auth issues... to be continued

hackingtype1:
sorry, unrelated error, reopening

hackingtype1:
found an issue where system thinks it is still scanning, but isn't. closing for the time being

PR #322, opened 2016-04-04
alpha 6

hackingtype1:


app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
```diff
@@ -83,42 +89,24 @@ public String getMenuName() {
...
-              } else {
-                  intent = new Intent(getApplicationContext(), Home.class);
-              }
-
-              startActivity(intent);
-              finish();
-          }
-
+            public void onClick(View v) {
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
AdrianLxM, line 99: I think this should only be done for G5.
It is missing for other InputMethods as well but they have a different logic for connection (-> permission request in bt scan activity).

hackingtype1, line 99: Yes, that sounds fine, we will update to only trigger if in G5 mode. Thanks!

PR #323, opened 2016-04-05
Save alerts data to preference during backup, so they can be restored…

tzachi-dar:
… from there.

Restoring will happen when we open xdrip, and after restoring we delete the key.


tzachi-dar:
One note about this solution.
It is possible that in the future the format of the alerts will change. In order to allow support for that, if no one has a better idea, I'll add a key with alerts version.

tzachi-dar:
Was merged in 327.

PR #325, opened 2016-04-21
Viewernightscoutmode

tzachi-dar:
This branch does 2 things:
1) There are fixes that are xDrip viewer related and allow it to read from a classic nightscout site. Classic means a site that the uploader to it is a nightscout uploader using a dexcom and an otg cable.

2) While working on this branch I have noticed that we don't set the hide_slope correctly so I have fixed that as well.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -95,11 +95,11 @@ android {
             resValue "string", "target_package", "com.eveningoutpost.dexdrip"
             resValue "string", "data_sync", "Data Sync"
         }
-        xdripviewer2 {
-            applicationId = "com.eveningoutpost.dexdrip.viewer2"
-            resValue "string", "content_provider", "com.example2"
-            resValue "string", "app_name", "xDripViewer22222222"
-            resValue "string", "target_package", "com.eveningoutpost.dexdrip.viewer2"
+        xdripviewer {
+            applicationId = "com.eveningoutpost.dexdrip.viewer"
+            resValue "string", "content_provider", "com.example1"
+            resValue "string", "app_name", "xDripViewer"
+            resValue "string", "target_package", "com.eveningoutpost.dexdrip.viewer"
             resValue "string", "data_sync", ""
         }
     }
@@ -107,8 +107,8 @@ android {
 
 dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
-//    xdripWearApp project(path: ':wear', configuration: 'xdripRelease')
-//    xdripviewerWearApp project(path: ':wear', configuration: 'xdripviewerRelease')
+    xdripWearApp project(path: ':wear', configuration: 'xdripRelease')
+    xdripviewerWearApp project(path: ':wear', configuration: 'xdripviewerRelease')
     testCompile 'com.squareup.okhttp:mockwebserver:2.6.0'
     compile 'com.nispok:snackbar:2.10.8'
     compile 'com.android.support:appcompat-v7:23.1.0'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -95,11 +95,11 @@ android {
             resValue "string", "target_package", "com.eveningoutpost.dexdrip"
             resValue "string", "data_sync", "Data Sync"
         }
-        xdripviewer {
-            applicationId = "com.eveningoutpost.dexdrip.viewer"
-            resValue "string", "content_provider", "com.example1"
-            resValue "string", "app_name", "xDripViewer"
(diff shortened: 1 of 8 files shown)


app/build.gradle
```diff
@@ -95,20 +95,20 @@ android {
...
         }
-        xdripviewer {
-            applicationId = "com.eveningoutpost.dexdrip.viewer"
-            resValue "string", "content_provider", "com.example1"
-            resValue "string", "app_name", "xDripViewer"
-            resValue "string", "target_package", "com.eveningoutpost.dexdrip.viewer"
+        xdripviewer2 {
+            applicationId = "com.eveningoutpost.dexdrip.viewer2"
+            resValue "string", "content_provider", "com.example2"
+            resValue "string", "app_name", "xDripViewer22222222"
```
AdrianLxM: These are just for testing?

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -240,16 +246,10 @@ public static void create(EGVRecord egvRecord, long addativeOffset, Context cont
                 bgReading.raw_calculated = (((calSlope * bgReading.raw_data) + calIntercept) - 5);
             }
             Log.i(TAG, "create: NEW VALUE CALCULATED AT: " + bgReading.calculated_value);
-            bgReading.calculated_value_slope = bgReading.slopefromName(egvRecord.getTrend().friendlyTrendName());
+            AtomicBoolean hide = new AtomicBoolean();
+            bgReading.calculated_value_slope = bgReading.slopefromName(egvRecord.getTrend().friendlyTrendName(), hide);
```
AdrianLxM: 'hide' looks like a parameter passed to the function but it actually is a return parameter. Is this a wanted side-effect?
I'd prefer it as real return value if is a return value. (class bundling a `double` and a `boolean`).

app/src/main/java/com/eveningoutpost/dexdrip/Services/XDripViewer.java
```diff
@@ -25,7 +27,11 @@
 //need synchronization
 
 public class XDripViewer extends AsyncTaskBase {
-    
+
+    static Sensor sensor_exists = null; // Will hold reference to any existing sensor, to avoid looking in DB
+    static Boolean isNightScoutMode = null;
+    public final static String NIGHTSCOUT_SENSOR_UUID = "c5f1999c-4ec5-449e-adad-3980b172b921";
```
AdrianLxM, line 33: What exactly is this String/ID? Is it always the same in certain cases? What cases are that? Only when xDrips sets it as it does not find a Sensor - or can it be set otherwise?

app/src/main/java/com/eveningoutpost/dexdrip/Services/XDripViewer.java
```diff
@@ -178,6 +190,84 @@ public static boolean isxDripViewerConfigured(Context ctx) {
...
         return true;
     }
+    
+    // This function checkes if we are looking at a nightscout site that is being uploaded by dexcom.
+    // In this case, we will not see sensors, and we need to create one.
+    // We identify such a mod by the fact that there are no sensors, and by the fact that on this bg,
+    // we see device == dexcom
+    // We make sure that all decisions will be cached.
+    
+    public static boolean IsNightScoutMode(Context context) {
```
AdrianLxM: `IsNightScoutMode` -> `isNightScoutMode`

app/src/main/java/com/eveningoutpost/dexdrip/Services/XDripViewer.java
tzachi-dar, line 33: In the case that we are working with a "classic nightscout uploader" there are no sensors. So practically speaking that means that we will never see a sensor. So in this case (only), I create one sensor and all packets will be using that.

app/src/main/java/com/eveningoutpost/dexdrip/Services/XDripViewer.java
tzachi-dar: Done.

app/build.gradle
tzachi-dar: Sure, fixed that.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
tzachi-dar: Passing variables by reference is very common on Cpp, but less in java. 
In any case, I have replaced this with returning a pair.

PR #335, opened 2016-04-29
alpha 8.1 - Receiver users only

hackingtype1:
I would tag this as "alpha 8.1 for receiver users only" in the group. The primary change is a 2 second connection delay; hopefully this makes the receiver + android combo more reliable. If it does, I will likely move this option to the G5 specific preferences.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/G5CollectionService.java
@@ -109,7 +109,6 @@ public class G5CollectionService extends Service {
     private long startTimeInterval = -1;
     private int lastBattery = 216;
     private Boolean isBondedOrBonding = false;
-    private Boolean isFirstTry = true;
 
     private AlarmManager alarm;// = (AlarmManager) getSystemService(ALARM_SERVICE);
 
@@ -187,7 +186,6 @@ public class G5CollectionService extends Service {
                     if (transmitterIdLastTwo.equals(deviceNameLastTwo)) {
                         isBondedOrBonding = true;
                     }
-
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Services/G5CollectionService.java
```diff
@@ -282,11 +277,9 @@ public void startScan() {
 
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
             setupLeScanCallback();
-
             mBluetoothAdapter.startLeScan(new UUID[]{ BluetoothServices.Advertisement }, mLeScanCallback);
         } else {
             Log.d(TAG, "startScan");
-
             mLEScanner.startScan(filters, settings, mScanCallback);
```
jamorham, line 283: Does this need a `catch` for `IllegalStateException` if Bluetooth is switched off? (Line 283 around startScan)

tzachi-dar, line 283: Well, at least according to the documentation, (https://developer.android.com/reference/android/bluetooth/le/BluetoothLeScanner.html#startScan(java.util.List<android.bluetooth.le.ScanFilter>, android.bluetooth.le.ScanSettings, android.bluetooth.le.ScanCallback) It does not seem needed.
Did you get such an exception?

jamorham, line 283: I think it is possible, Stephen previously referenced a similar issue:
https://gitter.im/nightscout/dexdrip/archives/2015/08/08

Line of code from Android source which raises the exception:
https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/bluetooth/le/BluetoothLeUtils.java#136

Called from startScan:
https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/bluetooth/le/BluetoothLeScanner.java#120

tzachi-dar, line 283: I guess that you are right, we need to catch that exception.

jstevensog, line 283: We should personally abstract out the BLE stuff, so an improvement in one will improve them all.  If it's possible that is.
Cheers

---- tzachi-dar wrote ----

PR #338, opened 2016-05-05
Limitter

AdrianLxM:
The LimiTTer connects to a Freestyle Libre sensor and emulates a BTWixel.
In some cases (age adjustment & calibration slope) it will use special parameters.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/BluetoothScan.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/BluetoothScan.java
@@ -309,7 +309,7 @@ public class BluetoothScan extends ListActivityWithMenu {
             if (!
                     (CollectionServiceStarter.isBTWixel(getApplicationContext())
                             || CollectionServiceStarter.isWifiandBTWixel(getApplicationContext())
-                    )) {
+                    ) || CollectionServiceStarter.isLimitter(getApplicationContext())) {
                 prefs.edit().putString("dex_collection_method", "BluetoothWixel").apply();
             }
             returnToHome();
--- a/app/src/main/java/com/eveningoutpost/dexdrip/BluetoothScan.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/BluetoothScan.java
@@ -313,6 +313,11 @@ public class BluetoothScan extends ListActivityWithMenu {
                 prefs.edit().putString("dex_collection_method", "BluetoothWixel").apply();
             }
             returnToHome();
+        } else if(device.getName().toLowerCase().contains("limitter")) {
+            if (!CollectionServiceStarter.isLimitter(getApplicationContext())) {
+                prefs.edit().putString("dex_collection_method", "LimiTTer").apply();
+            }
+            returnToHome();
         } else {
             returnToHome();
         }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -15,6 +15,7 @@ import com.eveningoutpost.dexdrip.ImportedLibraries.dexcom.records.EGVRecord;
 import com.eveningoutpost.dexdrip.ImportedLibraries.dexcom.records.SensorRecord;
 import com.eveningoutpost.dexdrip.ShareModels.ShareUploadableBg;
 import com.eveningoutpost.dexdrip.UtilityModels.BgSendQueue;
+import com.eveningoutpost.dexdrip.UtilityModels.CollectionServiceStarter;
 import com.eveningoutpost.dexdrip.UtilityModels.Constants;
 import com.eveningoutpost.dexdrip.UtilityModels.Notifications;
 import com.google.gson.Gson;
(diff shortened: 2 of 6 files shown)


tzachi-dar:
Generally looks good, a few comments inside.

app/src/main/java/com/eveningoutpost/dexdrip/BluetoothScan.java
```diff
@@ -309,10 +309,15 @@ protected void onListItemClick(ListView l, View v, int position, long id) {
             if (!
                     (CollectionServiceStarter.isBTWixel(getApplicationContext())
                             || CollectionServiceStarter.isWifiandBTWixel(getApplicationContext())
-                    )) {
+                    ) || CollectionServiceStarter.isLimitter(getApplicationContext())) {
```
tzachi-dar, line 312: If we have indeed seen drip in the name than I think we should change even if the user has chosen Limitter. 
This is probably a user mistake.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -709,9 +710,9 @@ public void find_new_curve() {
         }
     }
 
-    public void calculateAgeAdjustedRawValue(){
+    public void calculateAgeAdjustedRawValue(Context context){
```
tzachi-dar, line 713: please start the function with:
if(CollectionServiceStarter.isLimitter(context)) {
    age_adjusted_raw_value = raw_data;
    return;
}

it will make it much readable.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
```diff
@@ -747,5 +785,52 @@ public static double min_recent() {
                 .orderBy("timestamp desc")
                 .execute();
      }
-
 }
+
+abstract class SlopeParameters {
+    protected  double LOW_SLOPE_1 = 0.95;
```
tzachi-dar: There is no need to initialize the variables here.

app/src/main/res/values/arrays.xml
```diff
@@ -7,6 +7,7 @@
         <item>Wifi Wixel + BT Wixel</item>
         <item>G4 Share Receiver</item>
         <item>G5 Transmitter</item>
+        <item>LimiTTer</item>
```
tzachi-dar, line 10: Are we sure that the name that the user will see is LimiTTer?
Is there anyway to connect it with the libere?

app/src/main/java/com/eveningoutpost/dexdrip/BluetoothScan.java
AdrianLxM, line 312: That's what it does. The new part is outside of the negation (`!`).

app/src/main/res/values/arrays.xml
AdrianLxM, line 10: It is connected to an HM11. The name is set to "LimiTTer", yes.

AdrianLxM, line 10: You can see the project here: http://joernl.github.io/LimiTTer/

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
AdrianLxM: They will be initialized by the system anyways with system defaults - so not initializing them won't give any compiler warnings. I'd prefer to have meaningful defaults instead if someone misses a line. I assume a few more setups will follow once xDrip for enlite will be there.
I will default them to system defaults anyway if you really like that.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
AdrianLxM, line 713: You really dislike design principles like "single point of return", do you? ;)

AdrianLxM:
Ok, did some cleanup.

ghost:
Still works with great with XBridge2 setup in Wifi+Bt and XBridge Wixel settings. I appreciate the filtered values in the graph.
Thanks to all contributors

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
tzachi-dar, line 713: Single point of return is a good thing when it is combined with another keyword. Since this word is not used in java, it is a bad thing. Not specifying the keyword but you can see it here (www.javaworld.com/community/node/3375). Specifically picked a link that does not specify the word itself.

app/src/main/res/values/arrays.xml
tzachi-dar, line 10: a few questions:
1) If I understand correctly, http://joernl.github.io/LimiTTer/ does not contain any sw, is that correct?
2) Does the limiter use calibrations? If not we can probably remove them from the menus.

As for the name itself, shouldn't the name that the user is choosing be something like Freestyle?

app/src/main/java/com/eveningoutpost/dexdrip/BluetoothScan.java
tzachi-dar, line 312: You are right, my mistake.

app/src/main/res/values/arrays.xml
AdrianLxM, line 10: 1) Yes it does. Like the firmware for the wixel (wixel-xDrip or xBridge2) it has a firmware for the microcontroller that operates the NFC reader and the HM11.
2) It uses calibrations but at the moment just to set the offset (intercept). Once there is more data, we might adjust the slope as well (very likely, but not in as wide a range as for Dexcom sensors).

tzachi-dar:
Ok, I have no more comments, so merge at will.

PR #339, opened 2016-05-05
Viewernightscoutmode3

tzachi-dar:


tzachi-dar:
This is actually Viewernightscoutmode #325 merge.

PR #340, opened 2016-05-05
Add support for passing calculated filtered value to the xDrip viewer.

tzachi-dar:


AdrianLxM:
This looks straight forward to me. If the data comes through to xDripViewer, I'd say: merge!

thanks

PR #346, opened 2016-05-31
Xbridge and wifi working togeather.

tzachi-dar:


danpowell88:
All appears to be working correctly for me after a few days of testing.

tzachi-dar:
Good to know, so if there are no more comments, I'll merge that to master soon.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
```diff
@@ -34,11 +34,29 @@
...
  */
+
+class PointValueExtended extends PointValue {
+	
+	public PointValueExtended(float x, float y, float filtered) {
+		set(x, y);
+		calculatedFilteredValue = filtered;
+	}
+	   public PointValueExtended(float x, float y) {
+	        set(x, y);
```
AdrianLxM, line 50: Shouldn't that be a simple super-call?

tzachi-dar:
Do you means something like this?
public PointValueExtended(float x, float y) {
            super(x, y);
}

AdrianLxM:
Yes. Might at the moment be the exact same to call `super` instead of 
`set` (¿? - I didn't look at the constructor of the superclass) but 
would make it more robust to changes in the library in case they start 
doing something more in the constructor.

On 06/06/16 17:32, tzachi-dar wrote:

> Do you means something like this?
> public PointValueExtended(float x, float y) {
> super(x, y);
> }
> 
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub 
> https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/346#issuecomment-223995665, 
> or mute the thread 
> https://github.com/notifications/unsubscribe/AJPmwvtXmpCwbj88NxZA8ZLXRoa5_00Oks5qJD2hgaJpZM4IrDbU.

---

FREE 3D EARTH SCREENSAVER - Watch the Earth right on your desktop!
Check it out at http://www.inbox.com/earth

tzachi-dar:
Fixed in the DisableAlerts branch.

PR #350, opened 2016-06-08
Add timestamp to avoid pebble duplicate detection for BG alarms.

tanja3981:


AdrianLxM:
Looks pretty straight foreward to me. I think we can merge this quickly.
Danke @tanja3981.

BTW does the same thing (deduping by pebble) happen with missed reading alerts?

tanja3981:
I'm not using those, so I haven't tried. ;-)
The code change is for BG alerts only, so I guess I should add it there, too.

tzachi-dar:
Seems good to me.

tanja3981:
I added the time for the other notifications (calibration, signal lost, bg rise/fall), too.
Should be fine now.

AdrianLxM:
Danke @tanja3981

PR #364, opened 2016-07-26
No Broadcast Permission

AdrianLxM:
As agreed upon: No more permission.

Tested it with older NightWatch versions - works.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -2,8 +2,6 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.eveningoutpost.dexdrip" >
 
-    <permission android:name="com.eveningoutpost.dexdrip.permissions.RECEIVE_BG_ESTIMATE" />
-
     <!-- To receive data from HAPP. -->
     <uses-permission android:name="com.eveningoutpost.dexdrip.permissions.RECEIVE_EXTERNAL_STATUSLINE" />
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgEstimateBroadcaster.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgEstimateBroadcaster.java
@@ -1,0 +1,77 @@
+package com.eveningoutpost.dexdrip.UtilityModels;
+
+import android.content.Context;
+import android.content.Intent;
+import android.os.Bundle;
+
+/**
+ * Created by adrian on 26/07/16.
+ */
+public class BgEstimateBroadcaster {
+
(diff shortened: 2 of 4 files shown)


MilosKozak:
great

AdrianLxM:
I also refactored out the whole broadcast that it can be used by other apps (like the Medronic ones) to have the same kind of connectivity. Also tested on NightWatch.

MilosKozak:
do you broadcast xdrip version somehow to be able to warn user if using old xdrip (after i remove permission from nsclient)?

MilosKozak:
i got the idea now  but maybe now would be the right time to rename/add this broadcast to something else (like info,nightscout.sgv) and keep the old as it is now at least for some time

victorkp:
@MilosKozak if you leave the uses-permission in nsclient, then it should work with both old and new versions of xdrip. On Android, if you `<use-permission android:name="non-existant_permission"/>` then it's just a noop.

MilosKozak:
well there are at least 4-5 apps (as i know) using this broadcast. i think we must remove permission everywhere to get rid of issues we have now. that's why i'm thinking about new name

tzachi-dar:
Can we also pass the filtered bg value and filtered raw as well?

AdrianLxM:
@MilosKozak I can add a sender-ID String. Old NightWatch versions will just ignore them.

Any Idea for that?
Or Sender-ID-String and a Version-Number.

MilosKozak:
i'd prefer new broadcast as close as possible to NS record

AdrianLxM:
@MilosKozak I think that's a different thing. I'd really like to keep the legacy xDrip broadcast for now.

But if you want a new broadcast with new names, feel free to add/modify BgEstimateBroadcaster.java to what you need "as close as possible to NS record". I'm happy to add this to xDrip then.

MilosKozak:
@AdrianLxM  i know i was pushing on you to remove permission but now i see new broadcasts as better solution. it would make lot of things easier when we move something like NS API to internal communication

MilosKozak:
maybe broadcast JSONs sending to NS?

AdrianLxM:
I meant for (just?) xDrip I'd like to keep the legacy broadcast.

But before giving out library classes for other apps (even though it is not much): Now is the time to have a new/better broadcast format.

I actually don't mind the internal format as we should have one library for all apps. Would you want to define a format/structure that fits NS-Client as a central data hub, @MilosKozak?

I'd like to decouple this from this PR though.

MilosKozak:
ok open new thread for this discussion somewhere

AdrianLxM:
ah, just to add: If it stays within android, I don't see a reason to convert it to JSON (and then back on the receiving side).

PR #373, opened 2016-08-06
Take the correct time stamp even when find_new_raw_curve is not calle…

tzachi-dar:
…d on last packet.

Let's run with it and see if there are no problems.

Based on a bug found by JohanDegraeve.


tzachi-dar:
If there are no more remarks, I want to merge it soon.

PR #377, opened 2016-08-17
Remove duplicate settings

AdrianLxM:
They seem to have been duplicated during the merges.


tzachi-dar:
Indeed this looks duplicate.

PR #381, opened 2016-08-30
Mongo3.4

tzachi-dar:
This are the changes that are needed in order to run against mongo 3.2 driver (the next mlab forced update).
It is based on 3.4 snapshot drivers that also caused an update of ActiveAndroid to the latest version.

Any help in testing this is highly appreciated.

From my experience, current version of xdrip does work against 3.2 driver despite what mlab has published.


tzachi-dar:
This seems to be working well so far. If there are no more comments, than I will submit this soon, so that we are ready for the mlab update.

PR #383, opened 2016-08-31
Less distance between title and info

AdrianLxM:
The distance between the title/icon (action bar) and the info texts "minute ago" and BG value always has been quite large:
screenshot_2016-08-21-12-09-46

Now it looks like:
screenshot_2016-08-31-16-19-00

This is much space on a small phone that now can even be rotated:
screenshot_2016-08-31-16-20-22

Please test on larger phones! Thanks.

### DIFF
--- a/app/src/main/res/layout/activity_home.xml
+++ b/app/src/main/res/layout/activity_home.xml
@@ -3,18 +3,12 @@
     android:layout_width="match_parent" android:layout_height="match_parent"
     tools:context="com.eveningoutpost.dexdrip.Home">
 
-    <!-- As the main content view, the view below consumes the entire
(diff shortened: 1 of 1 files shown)


tzachi-dar:
Let me try this on the bigger tablets and see what I'll get.

tzachi-dar:
This looks well on a 10" tablet.
On the 7.5" tablet this has a small problem when a signal is missed. (The high part of the S is covered).

But the real problem is with the nexus 5x. Where we miss the last line all together.

screenshot_20160902-112315

Perhaps the solution is to make the changes for small devices only (and test what happens when a signal is missed there).

PR #385, opened 2016-09-08
models in manifest

AdrianLxM:


tzachi-dar:
This should probably fix the issue.

PR #402, opened 2017-03-10
Made it translateable

TebbeUbben:


AdrianLxM:
Hi, if you have updates to the translations or want to add a new language, please do so on the new repository: https://github.com/NightscoutFoundation/xDrip

TebbeUbben:
Ok, but why are calibration check-ins not working? I just used xDrip beta because they did not work in xDrip+.

